PR #477, opened 2015-10-29
Solve #456: Star profile view show FWHM in larger font when the view is higher th…

### LINKED ISSUE #456: Request to display FWHM in larger font for focusing
from Stark Labs yahoo group:

This is a request.  Could the FWHM text be shown in large text?  When focusing my guidescope I would like to be able to read the FWHM value shown on the star profile window.

Thanks,

Jeff

hholi:
…an half the systems screen.
Fixes #456 

When the star profile view is larger than half the systems screen the font for FWHM will grow as the height increases.
star profile normal
![star profile half screen height]
(https://cloud.githubusercontent.com/assets/7082679/10821007/69fa5aa4-7e50-11e5-8cc7-3efbd9b12e0a.png)
star profile full screen height

### DIFF
--- a/star_profile.cpp
+++ b/star_profile.cpp
@@ -151,7 +151,7 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
 
     int i;
     int *profptr;
-    string profileLabel;
+    wxString profileLabel;
     switch (this->mode) {  // Figure which profile to use
     case 0: // mid-row
     default:
--- a/star_profile.cpp
+++ b/star_profile.cpp
@@ -120,6 +120,30 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
     const int xsize = this->GetSize().GetX();
     const int ysize = this->GetSize().GetY();
 
+#if defined (__APPLE__)
+    dc.SetFont(*wxSMALL_FONT);
+#else
+    dc.SetFont(*wxSWISS_FONT);
+#endif
+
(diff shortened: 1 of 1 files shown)


star_profile.cpp
```diff
@@ -127,20 +151,20 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
...
     int i;
     int *profptr;
-    wxString label;
+    string profileLabel;
     switch (this->mode) {  // Figure which profile to use
     case 0: // mid-row
     default:
         profptr = midrow_profile;
-        label = _("Mid row");
+        profileLabel = "Mid row";
```
agalasso: _("Mid row") is important since this marks the string for translation (gettext).  We need to keep the _() !
_() is a macro that expands to wxGetTranslation() so if the user has selected a different language, "Mid row" is replaced by the corresponding localized version of the text.

hholi: I just pushed a correction to this.
Hallgeir

2015-10-29 16:39 GMT+01:00 Andy Galasso <notifications@github.com>:

star_profile.cpp
```diff
@@ -127,20 +151,20 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
 
     int i;
     int *profptr;
-    wxString label;
+    string profileLabel;
```
agalasso: using std::string will introduce some odd implicit conversions, and I'm not sure it will produce the right results for all locales.  Better to stick with wxString since a) the return type of _() is wxString, and b) the %s in wxString::Format() expects a wxString

Issue #480, opened 2015-10-30
Add HFD to star profile view for focusing


hholi:
Using the HFD in large font and making better use of screen area for the number.

agalasso:
fixed by Halgeir
b6dd190012293881dbbb72da0336381cfbcf0951

PR #481, opened 2015-10-30
480 hfd

hholi:
Solves #480 

Replaced FWHM number shown in window with HFD.
Rule for showing HFD in large font is now that view y > view x / 2 to make better use of screen area.


hholi:
Added back the FWHM label.

PR #515, opened 2016-03-03
issue 426.  added support for one shot move.  Mouse position captured…

johnson-cj:
… with ctrl-alt-mouseclick and position offset sent to mount.

### DIFF
--- a/guider_onestar.cpp
+++ b/guider_onestar.cpp
@@ -635,7 +635,27 @@ void GuiderOneStar::OnLClick(wxMouseEvent &mevent)
 {
     try
     {
-        if (mevent.GetModifiers() == wxMOD_CONTROL)
+        //single shot move -- C.Johnson 3/2/2016
+	if (mevent.GetModifiers() == wxMOD_ALTGR)
+        	{
+		//if no star selected or mount not connected, do nothing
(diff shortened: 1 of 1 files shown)


agalasso:
Hi CJ, Sorry for the delay getting back to you on this.
I think the code works as a good proof-of-concept, but I think there are some items that we'll need to address before merging it.

1. a large mount move in the main thread blocks the main thread (UI) for potentially many seconds. Perhaps the move should be done in the worker thread, not the main thread so the UI is not blocked, but there are implications for handling what can and cannot happen in the UI when the move is in progress (we'd have to think about that).
2. the code as written does not appear to account for camera angle (right?)
3. the keyboard shortcut is a bit obscure and could potentially be disruptive if a user accidentally alt-clicked when meaning to Shift-click (deselect) or Ctrl-click (bookmark).  Perhaps right-click with a pop-up menu with an option "nudge to here" would be safer?

Thanks,
Andy

d33psky:
Hey johnson-cj, what do you think about Andy's remarks ?

agalasso:
CJ,
Since this PR has been sitting idle for so long I'm going to close it.  You can submit another PR if you are interested in pursuing it some other time.

PR #524, opened 2016-03-27
SXMacLib: Better timeout and disconnect support

simonct:
Adopt timeout from Camera settings, report read pixels errors to caller and disconnect camera, tidy up code to get camera params


agalasso:
Thanks Simon!

PR #535, opened 2016-04-26
SXMacLib: Isolate linux changes to the Linux build for now, add prefi…

simonct:
…x to log messages


d33psky:
Thanks for the patch and sorry for breaking Mac support. Let's rename the file to match its contents better, something like SXlibUSB ?

PR #549, opened 2016-07-31
Added new eventmonitoring methods: get_mount_guiding_enabled and set_…

RayGralak:
…mount_guiding_enabled

### DIFF
--- a/event_server.cpp
+++ b/event_server.cpp
@@ -1489,6 +1489,39 @@ static void get_camera_binning(JObj& response, const json_value *params)
         response << jrpc_error(1, "camera not connected");
 }
 
+static void get_mount_guiding_enabled(JObj& response, const json_value *params)
+{
+	if (pMount)
+		response << jrpc_result(pMount->GetGuidingEnabled());
+	else
+		response << jrpc_error(1, "mount not defined");
+}
+
+static void set_mount_guiding_enabled(JObj& response, const json_value *params)
+{
+	Params p("enabled", params);
+	const json_value *val = p.param("enabled");
+	bool enable;
+	if (!val || !bool_param(val, &enable))
+	{
+		response << jrpc_error(JSONRPC_INVALID_PARAMS, "expected enabled boolean param");
+		return;
+	}
+
+	VERIFY_GUIDER(response);
+
+	if (pMount)
+	{
+		pMount->SetGuidingEnabled(enable);
+		if (pFrame && pFrame->pAdvancedDialog)
+			pFrame->pAdvancedDialog->LoadValues();
+		response << jrpc_result(0);
+	}
+	else
+		response << jrpc_error(1, "mount not defined");
+}
+
+
 static void dump_request(const wxSocketClient *cli, const json_value *req)
 {
     Debug.Write(wxString::Format("evsrv: cli %p request: %s\n", cli, json_format(req)));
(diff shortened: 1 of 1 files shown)


agalasso:
Ray,

Thanks for the pull request. My comments:

1. There is some ambiguity regarding AO vs. Mount.  In the UI the user can enable and disable AO and Mount guide output independently, but the new code will enable/disable the AO if it is present, and the mount otherwise.  This is probably ok assuming the intent is to disable guiding for PEMPro to measure PE, but if this is the intent we should probably change the method name to make the intent more clear.  Perhaps: enable_guide_output / get_guide_output_enabled ?

2. returning a JSON error when mount (or AO) is "None". This is a valid configuration, so I don't think it should return a json error. So far we have reserved error responses for hard errors like invalid JSON requests, internal PHD2 error and such. I think it would be ok to return false when pMount is null.

If we are going to have enable_guide_output return a JSON error if pMount is null, then we probably should add a new method to allow an app to see if a mount and/or AO is present before attempting to call enable_guide_output. I just added a new method "get_current_equipment" that can be used for this purpose.

3. VERIFY_GUIDER(response) - this can be omitted since the method is not referencing pFrame->pGuider hence no point in validating it

4. please convert hard tabs to 4 spaces

Thanks,
Andy

RayGralak:
Hi Andy,

I think I'm ok with those changes. I'm unclear if you are asking me to make those changes, or you would make them if you were to integrate the code into the main branch? If you are asking me to do them then I think I may want to add few more methods before making another pull request.

-Ray

agalasso:
>  I'm unclear if you are asking me to make those changes
yes, please go ahead and make the changes

RayGralak:
Hi Andy,

OK, no problem. I had a look again at the #2 issue you mentioned, and I thought I was following the recommended method of reporting an error, as the recently added "get_camera_binning" code does this:

```
static void get_camera_binning(JObj& response, const json_value *params)
{
    if (pCamera && pCamera->Connected)
    {
        int binning = pCamera->Binning;
        response << jrpc_result(binning);
    }
    else
        response << jrpc_error(1, "camera not connected");
}
```
While my code does this:

```
static void get_mount_guiding_enabled(JObj& response, const json_value *params)
{
	if (pMount)
		response << jrpc_result(pMount->GetGuidingEnabled());
	else
		response << jrpc_error(1, "mount not defined");
}
```
Just so I can understand how responses like this should be done, why is it OK to return a json error for get_camera_binning if a camera is not connected, but it is not OK for get_mount_guiding_enabled to return an error if no mount has been defined? 

Thanks,

-Ray

agalasso:
> why is it OK to return a json error for get_camera_binning if a camera is not connected, but it is not OK for get_mount_guiding_enabled to return an error if no mount has been defined?

I don't think there was any other option other than to return an error if binning is unknown. In the guiding enabled case it is plausible to state "false" when the mount is set to "None". If you feel strongly that you would prefer to have it return an error, then I will not object.

RayGralak:
Hi Andy,

I believe that "guiding enabled" can be false (or true), even when a mount is defined and PHD2 is **not** guiding. I feel there should be a distinction between the setting being disabled and not being able to return the setting's value, so I would like to keep that json error as is. I think there is value in hinting to the caller what needs to be done (i.e., define a mount or AO device in PHD2).

I'll commit the other changes you requested tonight when I get home. I already edited those changes this morning and this was the only issue that needed clarification.

Best regards,

-Ray

Issue #565, opened 2016-11-16
cam_QHY5.cpp file bug - maybe solved


astrovr:
Hi, I'm new on OpenPHDGuiding project so first of all congratulations to all contributors for this work!
I'm owner of a Magzero Mz5-m camera (QHY5 camera for italian market brand) that doesn't work well with PHD2 on Raspberry (ubuntu mate 16.04 OS): in particular calibration and guiding procedures fail (i use a 9x50 finder as guidescope and all calibration and guiding settings are correctly set).
After many test I've found the reason: a bug in cam_qhy5.cpp file.
The bug is here:

-----
`bool Camera_QHY5Class::ST4PulseGuideScope(int direction, int duration)
{
    int result = -1;
    int reg = 0;
    int32_t dur[2] = { -1, -1};

    duration /= 10;

    if (duration >= 255) duration = 254; // Max guide pulse is 2.54s -- 255 keeps it on always`

-----
with this settings calibration steps (and guiding corrections I suppose) are limited to 254 ms, a too short value for guiding expecially with short guidescopes (i.e. 9x50 finder that i use).
This bug is confirmed by testing the funcion "Manual guide" where even with long time pulses (1000 ms and longer) corresponds always a short noise of motor movements, compatible with a 254 ms correction pulse.
Also the calibration procedures confirm this: instead of about 12 steps, 40 or higher steps are necessary to perform (a bad) calibration.
I modify the cpp file as follows:

-----
` bool Camera_QHY5Class::ST4PulseGuideScope(int direction, int duration)
{
    int result = -1;
    int reg = 0;
    int32_t dur[2] = { -1, -1};

    // duration /= 10;

    if (duration >= 2500) duration = 2500; // Max guide pulse is 2.54s -- 255 keeps it on always `

----
and rebuild PHD2 according the procedure for Ubuntu.
Now calibration and guiding work. Calibration is done in about 12 steps and I don't lose the guiding star, even with a rough polar alignment.
In attachment the cam_qhy5.cpp file (you have to unzip the folder).
Sorry for my bad english...

Matteo

[cam_qhy5.zip](https://github.com/OpenPHDGuiding/phd2/files/594037/cam_qhy5.zip)

agalasso:
Matteo, great work, thanks for the fix!
The best way to get your change into the master branch would be for you to submit a new Pull Request.
Create your own fork containing your change, then submit the pull request.
Andy

astrovr:
[...]

PR #566, opened 2016-11-16
Bug fixed as issue #565 - cam_qhy5.cpp file

astrovr:
I'm owner of a Magzero Mz5-m camera (QHY5 camera for italian market brand) that doesn't work well with PHD2 on Raspberry (ubuntu mate 16.04 OS): in particular calibration and guiding procedures fail (i use a 9x50 finder as guidescope and all calibration and guiding settings are correctly set).
After many test I've found the reason: a bug in cam_qhy5.cpp file.
The bug is here:

`bool Camera_QHY5Class::ST4PulseGuideScope(int direction, int duration)
{
int result = -1;
int reg = 0;
int32_t dur[2] = { -1, -1};

duration /= 10;

if (duration >= 255) duration = 254; // Max guide pulse is 2.54s -- 255 keeps it on always`
with this settings calibration steps (and guiding corrections I suppose) are limited to 254 ms, a too short value for guiding expecially with short guidescopes (i.e. 9x50 finder that i use).
This bug is confirmed by testing the funcion "Manual guide" where even with long time pulses (1000 ms and longer) corresponds always a short noise of motor movements, compatible with a 254 ms correction pulse.
Also the calibration procedures confirm this: instead of about 12 steps, 40 or higher steps are necessary to perform (a bad) calibration.
I modify the cpp file as follows:

` bool Camera_QHY5Class::ST4PulseGuideScope(int direction, int duration)
{
int result = -1;
int reg = 0;
int32_t dur[2] = { -1, -1};

// duration /= 10;

if (duration >= 2500) duration = 2500; // Max guide pulse is 2.54s -- 255 keeps it on always `
and rebuild PHD2 according the procedure for Ubuntu.
Now calibration and guiding work. 

Matteo

### DIFF
--- a/cam_qhy5.cpp
+++ b/cam_qhy5.cpp
@@ -146,9 +146,9 @@ bool Camera_QHY5Class::ST4PulseGuideScope(int direction, int duration)
     int reg = 0;
     int32_t dur[2] = { -1, -1};
 
-    duration /= 10;
+    // duration /= 10;
 
-    if (duration >= 255) duration = 254; // Max guide pulse is 2.54s -- 255 keeps it on always
+    if (duration >= 2500) duration = 2500; // Max guide pulse is 2.54s -- 255 keeps it on always
     switch (direction)
     {
     case WEST:


agalasso:
merged, thanks!

PR #570, opened 2016-12-07
Properly use Info.plist with cmake, enable Retina support

markus-k:
This pull request removes the unused `PHD-Info.plist` (correct me if I'm wrong, couldn't find it in use anywhere) and replaces it a template from CMake, `MacOSXBundleInfo.plist.in`.

The new plist also has the key for HiDPI/Retina support set, which wasn't working beforehand.


agalasso:
good stuff, thanks!

PR #606, opened 2017-05-07
Allow usage of system's Eigen3 and Gtest

mattiaverga:
Follow up of previous pull request, now it's possible to use system's Eigen3 and Gtest also.

I moved the relevant options to _compiler_options.cmake_.


raffienficiaud:
Hi, why do we need to have the FindGtest.cmake ? Isn't the module provided by cmake enough?

mattiaverga:
You're right, forgive me as I'm new to cmake and I didn't know of that module...
I've modified the pull request.

PR #607, opened 2017-05-13
Fix linux locales

mattiaverga:
Fix locales and help installation under linux and add Italian translation to the available languages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -797,27 +797,19 @@ endif()
 
 ## Linux bundling of documentation + translation
 if(UNIX AND NOT APPLE)
-  # install language help files
-  set(locales_help     en_EN de_DE es_ES fr_FR it_IT ja_JP pl_PL ru_RU uk_UA zh_CN zh_TW)
-  set(locales_help_msg en    de    es    fr    it    ja    pl    ru    uk    zh_CN zh_TW)
-  list(LENGTH locales_help _len)
-  math(EXPR _len "${_len} - 1")
-  foreach(_index RANGE ${_len})
 
-    list(GET locales_help       ${_index} current_locale)
-    list(GET locales_help_msg   ${_index} current_locale_install_name)
-
+  foreach(current_locale IN LISTS all_locales)
     if("${current_locale}" STREQUAL "${default_locale}")
       set(destination share/phd2/)
-      set(install_options DESTINATION ${CMAKE_INSTALL_PREFIX}/share/phd2/)
+      set(install_help_options DESTINATION ${CMAKE_INSTALL_PREFIX}/share/phd2/)
     else()
       set(destination share/phd2/locale/${current_locale}/)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -798,8 +798,8 @@ endif()
 ## Linux bundling of documentation + translation
 if(UNIX AND NOT APPLE)
   # install language help files
-  set(locales_help     en_EN de_DE es_ES fr_FR ja_JP pl_PL ru_RU uk_UA zh_CN zh_TW)
-  set(locales_help_msg en    de    es    fr    ja    pl    ru    uk    zh_CN zh_TW)
+  set(locales_help     en_EN de_DE es_ES fr_FR it_IT ja_JP pl_PL ru_RU uk_UA zh_CN zh_TW)
+  set(locales_help_msg en    de    es    fr    it    ja    pl    ru    uk    zh_CN zh_TW)
```
raffienficiaud: there is the list ``all_locales`` that is returned from the call to ``generate_doc_targets``. Do you think it would be possible to link those two lists above to that ``all_locales`` list? There is the special case of ``zh`` that I did not know how to handle...

mattiaverga:
FYI: at the moment locales installation on linux does not work at all: at line 833 the `${message_file_basename}` does not exist at all, so the If statement is never true, and at line 840 the `if(${current_zip})` doesn't seem to work (at least for me under Fedora).

raffienficiaud:
I understand that you are fixing it, but before I made the changes to the way locales are working on Linux, they were maybe working. I obviously introduced something there (see commit 85f220db7c18661e544efab1c45005b8fe7e1513), but I suggest we concentrate on fixing it once for all. 

As I said, my concern is about the way the folders are named in your changes. Your implementation is not doing the same as "the intent" of the current code and the implementation of the previous code (before the merge 85f220db7c18661e544efab1c45005b8fe7e1513). It was made this way for some reasons I do not know about.

Do not get me wrong, I am not saying that what you did is incorrect, it is different and I do not know who tested it and on what platform.

Thanks

pchev:
I look at my old ppa buildlog and I can confirm that the locale where installed correctly in April 11 version but not in April 29 version. 

I test Mattia branch fix_linux_locales and I confirm it work on my Ubuntu system.

The only problem by installing the locale using this directory name is with the "system default" language selection if you use another language variant like fr_BE or fr_CH. 
But this only affect the "system default" language selection, if you select "French" in the brain it work in all case.

agalasso:
Great, thanks for testing it Patrick.
Unless anyone has any objections, I'll merge it shortly.

pchev:
For me you can merge this change.

I will rebuild my ppa version soon after as the current one miss the translations and help files.

agalasso:
Mattia, thanks for the PR and sorry for the delay in merging. I was travelling and out of town.

mattiaverga:
@agalasso no problem, I'm a bit under pressure too in these days.
Thanks for accepting the PR.

PR #614, opened 2017-06-11
Add appdata file

mattiaverga:
Add appdata file in linux installations.

See https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps

### DIFF
--- a/phd2.appdata.xml
+++ b/phd2.appdata.xml
@@ -4,7 +4,7 @@
     <id>phd2.desktop</id>
     <metadata_license>CC0-1.0</metadata_license>
     <project_license>BSD-3-Clause</project_license>
-    <name>Comical</name>
+    <name>PHD2</name>
     <summary>Telescope guiding software</summary>
     
     <description>
@@ -17,7 +17,6 @@
             <li>Easy-to-use, “push here dummy” guiding for beginners</li>
             <li>Sophisticated guiding and analysis tools for experienced users</li>
             <li>Extensive support for commonly-used equipment</li>
-            <li>Easy upgrade/co-installation with PHD1, including many new features</li>
             <li>Extensive online help and responsive product support</li>
         </ul>
     </description>
@@ -28,10 +27,6 @@
         <screenshot type="default">
             <caption>The main window</caption>
             <image>http://openphdguiding.org/wp-content/uploads/2015/06/capture_21.jpg</image>
-        </screenshot>
-        <screenshot>
-            <caption>The log viewer</caption>
-            <image>http://openphdguiding.org/wp-content/uploads/2015/05/phdlogviewer_sm.png</image>
         </screenshot>
     </screenshots>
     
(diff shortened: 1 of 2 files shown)


phd2.appdata.xml
```diff
@@ -0,0 +1,40 @@
...
+        <p>
+        PHD2 is telescope guiding software that simplifies the process
+        of tracking a guide star, letting you concentrate on other aspects of
+        deep-sky imaging or spectroscopy.
+        </p>
+        <ul>
+            <li>Easy-to-use, “push here dummy” guiding for beginners</li>
+            <li>Sophisticated guiding and analysis tools for experienced users</li>
+            <li>Extensive support for commonly-used equipment</li>
+            <li>Easy upgrade/co-installation with PHD1, including many new features</li>
```
d33psky: This line makes sense for Windows (and maybe Mac) but no sense for Linux

phd2.appdata.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2017 Mattia Verga <mattia.verga@tiscali.it> -->
+<component type="desktop-application">
+    <id>phd2.desktop</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>BSD-3-Clause</project_license>
```
d33psky: Is this the license we want ?

phd2.appdata.xml
```diff
@@ -0,0 +1,40 @@
...
+    <name>Comical</name>
```
d33psky: What/who is 'comical' ?

phd2.appdata.xml
```diff
@@ -0,0 +1,40 @@
...
+    <launchable type="desktop-id">phd2.desktop</launchable>
+    
+    <screenshots>
+        <screenshot type="default">
+            <caption>The main window</caption>
+            <image>http://openphdguiding.org/wp-content/uploads/2015/06/capture_21.jpg</image>
+        </screenshot>
+        <screenshot>
+            <caption>The log viewer</caption>
+            <image>http://openphdguiding.org/wp-content/uploads/2015/05/phdlogviewer_sm.png</image>
```
d33psky: Do we have the logviewer for Mac and Linux ?

phd2.appdata.xml
agalasso: personally I think we should switch to GPL, but that will require approval from all the contributors who have put their names on license text throughout the code

phd2.appdata.xml
agalasso: yes: http://adgsoftware.com/phd2utils/ but we only have Patrick's PPA Ubuntu pre-built .deb, other distros need to build from source

phd2.appdata.xml
mattiaverga: The LICENSE.txt file distributed in sources is a BSD-3-Clause, so I put that in <project_license>.

The <metadata_license> only refers to the appdata file itself.

phd2.appdata.xml
mattiaverga: My fault... a copy/paste error 😮 Should be PHD2.

phd2.appdata.xml
mattiaverga: Ah, I thought it was included in the main PHD2 program... I will remove this screenshot.

PR #622, opened 2017-08-12
Direct attach SBIG in Linux

CheyenneWills:
Following changes:
- fix a link error for indilib (requires nova libraries)
- use "system" installed copy of SBIG SDK.
- color support for sbig

Fixed a load error with current indilib (indilib appears to need the nova libraries).

The changes are fairly small (mostly in thirdparty/thirdparty.cmake).
I created a section to handle the case where the sbig libraries are loaded on the system

Added check for color on sbig cameras and call the QuickLRecon function

(sorry about the line end change change in cam_SBIG.cpp for the color support.
Here's the diff ignoring whitespace for the color support)

```
@@ -70,6 +70,7 @@ Camera_SBIGClass::Camera_SBIGClass()
     UseTrackingCCD = false;
     HasShutter = true;
     HasSubframes = true;
+    IsColor = false;
 }
 
 Camera_SBIGClass::~Camera_SBIGClass()
@@ -348,6 +349,8 @@ bool Camera_SBIGClass::Connect(const wxString& camId)
     Debug.Write(wxString::Format("SBIG: %s type=%u, UseTrackingCCD=%d, MaxBin = %hu, 1x1 size %d x %d, 2x2 size %d x %d\n", gcir0.name, gcir0.cameraType, UseTrackingCCD, MaxBinning, m_imageSize[0].x, m_imageSize[0].y, m_imageSize[1].x, m_imageSize[1].y));
 
     Name = gcir0.name;
+    IsColor = Name.find("Color");
+   
     Connected = true;
     return false;
 }
@@ -544,6 +547,9 @@ bool Camera_SBIGClass::Capture(int duration, usImage& img, int options, const wx
     }
 
     if (options & CAPTURE_SUBTRACT_DARK) SubtractDark(img);
+    if (IsColor && Binning == 1 && (options & CAPTURE_RECON)) {
+        QuickLRecon(img);
+    }
 
     return false;
 }
```

### DIFF
--- a/cam_SBIG.cpp
+++ b/cam_SBIG.cpp
@@ -70,6 +70,7 @@ Camera_SBIGClass::Camera_SBIGClass()
     UseTrackingCCD = false;
     HasShutter = true;
     HasSubframes = true;
+    IsColor = false;
 }
 
 Camera_SBIGClass::~Camera_SBIGClass()
@@ -348,6 +349,8 @@ bool Camera_SBIGClass::Connect(const wxString& camId)
     Debug.Write(wxString::Format("SBIG: %s type=%u, UseTrackingCCD=%d, MaxBin = %hu, 1x1 size %d x %d, 2x2 size %d x %d\n", gcir0.name, gcir0.cameraType, UseTrackingCCD, MaxBinning, m_imageSize[0].x, m_imageSize[0].y, m_imageSize[1].x, m_imageSize[1].y));
 
     Name = gcir0.name;
+    IsColor = Name.find("Color");
+   
     Connected = true;
     return false;
 }
@@ -544,6 +547,9 @@ bool Camera_SBIGClass::Capture(int duration, usImage& img, int options, const wx
     }
 
     if (options & CAPTURE_SUBTRACT_DARK) SubtractDark(img);
+    if (IsColor && Binning == 1 && (options & CAPTURE_RECON)) {
+        QuickLRecon(img);
+    }
 
     return false;
 }
--- a/cam_SBIG.h
+++ b/cam_SBIG.h
@@ -38,6 +38,8 @@
 
 #if defined (__APPLE__)
 #include <SBIGUDrv/sbigudrv.h>
+#elif defined(__LINUX__)
+    #include <sbigudrv.h>
 #else
 #include "cameras/Sbigudrv.h"
 #endif
@@ -48,6 +50,7 @@ class Camera_SBIGClass : public GuideCamera
     bool m_driverLoaded;
     wxSize m_imageSize[2]; // 0=>bin1, 1=>bin2
     double m_devicePixelSize;
+    bool IsColor;
 
 public:
     Camera_SBIGClass();
--- a/cameras.h
+++ b/cameras.h
@@ -111,6 +111,7 @@
 # define INDI_CAMERA
 # define ZWO_ASI
 # define SXV
+# define SBIG
 
 #endif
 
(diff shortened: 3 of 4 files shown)


agalasso:
Thanks for the PR.  The latest INDI lib no longer depends on libnova, so I think that part should be removed. Perhaps your distro has not received that change from upstream INDI?

agalasso:
would it be possible to revert the line-ending changes in cam_SBIG.cpp so we can see the actual diff? (you can use unix2dos to get back to the original DOS-style line endings.)  This is our problem in the phd2 source tree since we really should not have the files saved with DOS line terminators, but until we fix this it would help if you could go with the flow, thanks.

CheyenneWills:
I did include the diff output  for cam_SBIG.cpp in the pull request comments above.  There are three small changes in cam_SBIG.cpp to add the check for color (create the flag, test the name returned for Color and then check to see if QuickLRecon needs to be called). 

(I can also do a unix2dos and commit that..  -- I understand the grief :) -- just let me know.)

I'm at indilib 1.4.1 (which is the current stable level level according to the indilib.org code repository).  I'm using a Gentoo system with an ebuild from -> http://indilib.org/download/gentoo.html.  The indilib source does require libnova according to their install instructions -> https://github.com/indilib/indi/blob/master/libindi/INSTALL

agalasso:
> I can also do a unix2dos and commit that

yes, please do

> I'm at indilib 1.4.1

the indilib changes that remove the libnova dependency came in on 7/19/2017, so they are not present in 1.4.1.  You'll need to use newer indilib code, so probably best to [build indilib from source](https://github.com/indilib/indi#building)

I updated the [PHD2 build instructions](https://github.com/OpenPHDGuiding/phd2/wiki/BuildingPHD2OnLinux) to clarify this... sorry you ran into the problem.

CheyenneWills:
I created a new branch and pushed the crlf change back into that (and created a new pull request off that).  (I have some local changes in the phd2 that is specific to a camera that has a "broken" firmware -- working with the vendor to fix)

I pulled the current indilib source and the current development level still has references to libnova. 
(not fully complete.. but..)
```
libindi/libs/indibase/indiccd.cpp:#include <libnova/julian_day.h>
libindi/libs/indibase/indiccd.cpp:#include <libnova/ln_types.h>
libindi/libs/indibase/indiccd.cpp:#include <libnova/precession.h>
libindi/libs/indibase/indidetector.cpp:#include <libnova/julian_day.h>
libindi/libs/indibase/indidetector.cpp:#include <libnova/ln_types.h>
libindi/libs/indibase/indidetector.cpp:#include <libnova/precession.h>
libindi/libs/indibase/indidome.cpp:#include <libnova/julian_day.h>
libindi/libs/indibase/indidome.cpp:#include <libnova/sidereal_time.h>
libindi/libs/indibase/indidome.cpp:#include <libnova/transform.h>
libindi/libs/indibase/indidome.h:#include <libnova/ln_types.h>
libindi/libs/indibase/inditelescope.h:#include <libnova/julian_day.h>
libindi/libs/indicom.c:#include <libnova/julian_day.h>
libindi/libs/indicom.c:#include <libnova/sidereal_time.h>
libindi/libs/indicom.h:    \param iso_date a pointer to a \e ln_date structure to store the extracted time and date (libnova).
```

agalasso:
> I pulled the current indilib source and the current development level still has references to libnova

right, when you build indi it will produce several libraries, some of which require libnova.  The indi client library (libindiclient.a) is the only one that gets linked with phd2 and libindiclient.a does not have any libnova dependency.

When you do the full indi build, it will require libnova, but the dependency will not propagate to phd2.

CheyenneWills:
Okay... I just pushed my changes.  latest removes the libnova update.

So.. the only real changes are adding color camera detecting in cam_SBIG and pulling in the SBIG universal drivers when loaded into the system.

agalasso:
closing this PR, following-up in PR #623

PR #623, opened 2017-08-13
Open phd guiding base

CheyenneWills:
o Add color support for cam_SBIG
o Latest stable indilib still requires nova
o Use a system installed level of SBIG universal drivers

### DIFF
--- a/cam_SBIG.cpp
+++ b/cam_SBIG.cpp
@@ -70,6 +70,7 @@ Camera_SBIGClass::Camera_SBIGClass()
     UseTrackingCCD = false;
     HasShutter = true;
     HasSubframes = true;
+    IsColor = false;
 }
 
 Camera_SBIGClass::~Camera_SBIGClass()
@@ -348,6 +349,8 @@ bool Camera_SBIGClass::Connect(const wxString& camId)
     Debug.Write(wxString::Format("SBIG: %s type=%u, UseTrackingCCD=%d, MaxBin = %hu, 1x1 size %d x %d, 2x2 size %d x %d\n", gcir0.name, gcir0.cameraType, UseTrackingCCD, MaxBinning, m_imageSize[0].x, m_imageSize[0].y, m_imageSize[1].x, m_imageSize[1].y));
 
     Name = gcir0.name;
+    IsColor = (Name.find("Color") != string::npos); 
+   
     Connected = true;
     return false;
 }
@@ -544,6 +547,9 @@ bool Camera_SBIGClass::Capture(int duration, usImage& img, int options, const wx
     }
 
     if (options & CAPTURE_SUBTRACT_DARK) SubtractDark(img);
+    if (IsColor && Binning == 1 && (options & CAPTURE_RECON)) {
+        QuickLRecon(img);
+    }
 
     return false;
 }
--- a/cam_SBIG.h
+++ b/cam_SBIG.h
@@ -38,6 +38,8 @@
 
 #if defined (__APPLE__)
 #include <SBIGUDrv/sbigudrv.h>
+#elif defined(__LINUX__)
+    #include <sbigudrv.h>
 #else
 #include "cameras/Sbigudrv.h"
 #endif
@@ -48,6 +50,7 @@ class Camera_SBIGClass : public GuideCamera
     bool m_driverLoaded;
     wxSize m_imageSize[2]; // 0=>bin1, 1=>bin2
     double m_devicePixelSize;
+    bool IsColor;
 
 public:
     Camera_SBIGClass();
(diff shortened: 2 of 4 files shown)


CheyenneWills:
I believe that this is what you wanted.. (Again sorry for some git noobness..)

agalasso:
ok, thanks, getting close now.

It looks like we have one issue to resolve: attempting to build PHD2 *without* the SBIG dev kit installed fails with a compile error:

```
In file included from /home/andy/projects/phd2/cam_SBIG.cpp:40:0:   
/home/andy/projects/phd2/cam_SBIG.h:42:26: fatal error: sbigudrv.h: No such file or directory   
     #include <sbigudrv.h>   
                          ^   
compilation terminated.   
CMakeFiles/phd2.dir/build.make:1081: recipe for target 'CMakeFiles/phd2.dir/cam_SBIG.cpp.o' failed   
```

So we have to decide whether the SBIG dev kit install is going to be mandatory or optional for building phd2 on linux. If we decide to make it mandatory, we could bundle the SBIG dev kit and update the cmake files to unpack it as we do with cfitsio and eigen. Otherwise, we'll need to update the linux build instructions and make sure the build bots are updated.  If we want to make the SBIG support optional, then we'll need to add some conditional compilation in cameras.h like we do with QHY_CAMERA (see the setting and use of HAVE_QHY_CAMERA in cameras.h and thridparty.cmake)

What do you think?

CheyenneWills:
I would say make it optional. 

I will look at how the QHY camera is dealt with and try to replicate it with the SBIG support.  I won't be able to look at till after the 21st (got an eclipse to deal with <grin>)

My comment it would be nice to have a general way of selecting which cameras are included (even making the indilib support as an "option").   I would also try to make the build process look first for system installed 3rd party dependencies before using an "included" library. (I can try to help out with some of this if you would like).


On another side comment... (covers openPHD as well as some other packages)
One of the problems that I've been dealing with is having several different packages all trying to pull in their own copy of libraries when trying to build.  In general build processes really should look to see if there is an installed copy of the library/dependency on the system before using a "supplied" copy.  I've been having to patch build scripts because the package supplies some outdated 3rd party library.  From a security standpoint this is a nightmare (not keeping up to date with upstream maintenance -- in real life I'm an oldschool systems programmer that deals with security and system administration tools, so I'm quite familiar with packaging and maintenance).

agalasso:
> I would say make it optional.

ok, I'll buy that

> I will look at how the QHY camera is dealt with and try to replicate it with the SBIG support. I won't be able to look at till after the 21st (got an eclipse to deal with )

I can help with this too in the next few days.  If you want, we can merge your PR with the SBIG commented-out in cameras.h so no build breakage, then we can add the conditional compilation based on whether the cmake file finds the SBIG dev kit.

If you'd like to go that route, just update the PR to comment-out the SBIG in cameras.h and I will merge it and take care of the cmake changes.  Otherwise, I'll wait until you have a chance to make the cmake file changes in the PR.

CheyenneWills:
I'd like to just wait until sometime later next week if that's okay.  (I'm focusing on getting ready for the eclipse on the 21st..)

CheyenneWills:
Okay -- I think the latest commit should do what you want.  Let me know.  (BTW the total eclipse was awesome) 

Build instructions:

- Install SBIG drivers onto the system.  (
- cmake -DSYSTEM_SBIG=1 ..
- make


Assumes SBIG's Universal driver has been loaded into the system and placed in a standard path. (e.g. sbigudrv.h in /usr/include, libsbigudrv.so in /usr/lib )
  
 SDK for linux can be found here -> ftp://ftp.diffractionlimited.com/pub/devsw/LinuxDevKit.tar.gz
  
 To rebuild libSBIG
 ```
cd ${WORKDIR}/LinuxDevKit/x86/c/testapp
 local sharedlink="-shared -Wl,-soname,libSBIG-1.33.0"
 g++ -c ${CXXFLAGS} -c -fPIC -I /usr/include/libusb-1.0 -I ${WORKDIR}/LinuxDevKit/x86/c/testapp csbigimg.cpp csbigcam.cpp
 g++ -L ${WORKDIR}/LinuxDevKit/x86/c/lib64/ ${sharedlink} -o libSBIG=1.33.0 csbigimg.o csbigcam.o -lm -lsbigudrv -lusb -lcfitsio
 ar -cvq libSBIG.a csbigimg.o csbigcam.o
```

CheyenneWills:
Oh.. Okay... wasn't sure on that.  -- I'll make the change

CheyenneWills:
I don't have a Mac or Windows system to try to build against.  Let me know if the include statements need tweaked in any way.

PR #651, opened 2017-10-29
Staticpa enhancements

ken-self:
Allow the user to center the helper image on the selected reference star. Plus some minor fixes and revision to the tutorial to match the latest screen layouts and operation.


staticpa_toolwin.cpp
```diff
@@ -1078,7 +1115,7 @@ bool StaticPaToolWin::RotateMount()
             else if (s_reqRot > 45)
             {
                 Debug.AddLine(wxString::Format("StaticPA: Pos#2 Too close to CoR actoffsetdeg=%.1f s_reqRot=%.1f", actoffsetdeg, s_reqRot));
-                return RotateFail(wxString::Format("Star is too close to CoR %.1f deg", actoffsetdeg ));
+                return RotateFail(wxString::Format("Star is too close to CoR (%.1f deg) - try another reference star", actoffsetdeg ));
```
agalasso: should tag this string for translation:
return RotateFail(wxString::Format(_("Star is too close ...

PR #652, opened 2017-10-29
Polardrift - prototype

ken-self:
Based on King's and Towne's papers. Not yet tested live but works in simulation. The only thing I am not certain of is the direction of the adjustments. The algorithm is very simple. With guiding turned off, the drift of the guide star is tracked as dx/dt and dy/dt by linear regression. These are used to calculate the magnitude and direction of the drift which in turn translates to magnitude and direction of adjustment needed. The adjustment is plotted on the main display as an overlay.


myframe.cpp
```diff
@@ -498,6 +502,7 @@ void MyFrame::SetupMenuBar(void)
     tools_menu->Append(MENU_STARCROSS_TEST, _("Star-Cross Test"), _("Run a star-cross test for mount diagnostics"));
     tools_menu->Append(MENU_GUIDING_ASSISTANT, _("&Guiding Assistant"), _("Run the Guiding Assistant"));
     tools_menu->Append(MENU_DRIFTTOOL, _("&Drift Align"), _("Run the Drift Alignment tool"));
+    tools_menu->Append(MENU_POLARDRIFTTOOL, _("&Polar Drift Align"), _("Run the Polar Drift Alignment tool"));
```
agalasso: These names are too confusing now for a new user.  Maybe the names are ok, but I think we need to improve the tool-tips to explain a bit more what the differences are. Perhaps something like this (name / tool-tip):
Drift Align / alignment of the mount's polar axis by minimizing declination drift
Polar Drift Align / drift alignment using analysis of star motion with scope pointed towards the pole
Static Polar Align / alignment of the mount's polar axis using rotation about the RA axis while pointed towards the pole

polardrift_toolwin.cpp
```diff
@@ -0,0 +1,392 @@
...
+    PolarDriftToolWin *win = static_cast<PolarDriftToolWin *>(pFrame->pPolarDriftTool);
+    return win->PaintHelper(dc, scale);
+}
+
+PolarDriftToolWin::PolarDriftToolWin()
+: wxFrame(pFrame, wxID_ANY, _("Polar Drift Alignment"), wxDefaultPosition, wxDefaultSize,
+wxCAPTION | wxCLOSE_BOX | wxMINIMIZE_BOX | wxSYSTEM_MENU | wxTAB_TRAVERSAL | wxFRAME_FLOAT_ON_PARENT | wxFRAME_NO_TASKBAR)
+{
+    t0 = 0;
+    sumt = sumt2 = sumx = sumx2 = sumy = sumy2 = sumtx = sumty = sumxy = 0.0;
```
agalasso: can we use an m_ prefix for member variables?

polardrift_toolwin.cpp
```diff
@@ -0,0 +1,392 @@
...
+    m_guideOutputDisabled = true;
+    num = 0;
+    s_drifting = true;
+    FillPanel();
+    return;
+}
+
+void PolarDriftToolWin::OnCloseBtn(wxCommandEvent& evt)
+{
+    // save the window position
```
agalasso: this could call OnClose() to eliminate the duplicated code

polardrift_toolwin.cpp
```diff
@@ -0,0 +1,392 @@
...
+    double xslope = (num*sumtx - sumt*sumx) / (num*sumt2 - sumt*sumt);
+    double yslope = (num*sumty - sumt*sumy) / (num*sumt2 - sumt*sumt);
+
+    double theta = degrees(atan2(yslope, xslope));
+    // In the northern hemisphere the star rotates clockwise, in the southern hemisphere anti-clockwise
+    // In NH the pole is to the right of the drift vector (-90 degrees) however in pixel terms (Y +ve down) it is to the left (+90 degrees)
+    // So we multiply by a_hemi to get the correct direction
+
+    alpha = theta + a_hemi * 90; // direction to the pole
+    offset = sqrt(xslope*xslope + yslope*yslope)*factor;  //polar alignment error in pixels
```
agalasso: `sqrt(xslope*xslope + yslope*yslope)` === `hypot(xslope, yslope)`

polardrift_toolwin.cpp
```diff
@@ -0,0 +1,392 @@
...
+    target = PHD_Point(current.X + offset*cos(radians(alpha)), current.Y + offset*(sin(radians(alpha))));
+
+    Debug.AddLine(wxString::Format("Polar Drift: num %d t0 %.1f tnow %.1f Pos: %.1f,%.1f PA err(px): %.1f Angle (deg): %.1f", num, t0, tnow));
+    Debug.AddLine(wxString::Format("Polar Drift: slopex %.1f slopey %.1f offset %.1f theta %1f alpha %.1f", xslope, yslope, offset, theta, alpha));
+    SetStatusText(wxString::Format("Time %.1f PA err(arcmin): %.1f Angle (deg): %.1f", tnow, offset*g_pxScale/60, norm(-alpha,-180,180)));
```
agalasso: tag for translation _(" ... ")

polardrift_toolwin.h
```diff
@@ -0,0 +1,110 @@
...
+    wxString c_instr;
+
+    enum PolarDriftCtrlIds
+    {
+        ID_HEMI = 10001,
+        ID_START,
+        ID_CLOSE,
+    };
+
+    bool g_canSlew;     // Mount can slew
```
agalasso: I'm not aware of the w_ / c_ / g_ / a_  prefix convention.  Could we go with the convention used in the rest of the code base:
m_  = member variable
s_ = static (global) variable (though not used consistently in the code)
no prefix for parameters or local variables

PR #656, opened 2017-10-31
Staticpa - code cleanup

ken-self:
Code cleanup to standardise on "m_" prefix and tag for text translation.
Use hypot function where appropriate
Reduce size of SPA window by toggling between instructions and star map.


agalasso:
looks good, thanks

PR #659, opened 2017-11-05
Fix some compiler warnings

mattiaverga:
These changes fix some compiler warnings, mostly comparisons between signed and unsigned.

In eegg.cpp file it seems that there's an if statement wrongly bracketed (or is the indentation wrong?).


d33psky:
looks all fine to me, but what compiler complains about these exactly ?

mattiaverga:
GCC7 on Fedora.
There are even more warnings about variables and functions which are set but not used, but I'm not expert enough to try to fix them. Here it is the full build log:
https://copr-be.cloud.fedoraproject.org/results/mattia/Astronomy/fedora-rawhide-x86_64/00658287-phd2/build.log.gz

PR #662, opened 2017-11-16
Staticpa updates

ken-self:
Brings th etutorial up to date and adds minor functional changes


agalasso:
fyi, you can avoid the merge commits (noise) if you use `pull -r`

PR #680, opened 2018-02-17
Polardrift restore guiding status

ken-self:
When PolarDrift tool is stopped restore guiding enablement to the state prior to PolarDrift. Also some fixes to debug messages and corrections to inline comments


agalasso:
Thanks Ken!

PR #682, opened 2018-03-31
Backlash compensation improvements

agalasso:
- improved backlash compensation dynamic adjustment
- add additional control options for backlash comp
- apply backlash comp for "fast recovery" moves after dithering
- apply backlash comp for AO mount bumps in declination


bwdev01:
Thanks, Andy.
 
  _____

PR #702, opened 2018-07-04
Configurable Calibration Distance

alexhelms:
This PR removes the hard coded calibration distance and allows for the distance to be configurable. This was explored as part of troubleshooting a Mach1 (https://www.cloudynights.com/topic/622947-mach1-dec-oscillations-our-experience/?view=getnewpost) and also provides users more flexibility in configuring calibration to better match their particular imaging setup. Stock calibration worked great at 952mm focal length, but at 2800mm it was inadequate given my image scale.

**New Features**
- Removed the hard coded limit (25 px or 5% of height of image, whichever is smaller)
- Added UI in Calibration Step Calculator
- Added UI in the Advanced Setup -> Guiding tab
- Added calibration distance to log

I have tested it on a Mach1 with a C11 EdgeHD with an OAG at a pixel scale of 0.35 arc-sec/px and I get a significantly better calibration. The current PHD version (2.6.5dev2) gave me consistently bad and inaccurate calibration results due to calibration not moving far enough. 25 pixels at that scale is simply not enough to get an accurate calibration.

I have attached some logs from before this PR and also some logs of me experimenting with different values for calibration distance. These logs are in chronological order with the last calibration being very good.

**New UI on Advanced Setup**
new advanced setup

**New UI on Calibration Step Calculator**
new step calculator

**Successful Calibration with New Calibration Distance**
Good PHD Calibration

**Logs**
[PHD2_GuideLog_2018-07-02_225736.txt](https://github.com/OpenPHDGuiding/phd2/files/2164164/PHD2_GuideLog_2018-07-02_225736.txt)
[PHD2_GuideLog_2018-07-03_202916.txt](https://github.com/OpenPHDGuiding/phd2/files/2164165/PHD2_GuideLog_2018-07-03_202916.txt)
[PHD2_GuideLog_2018-07-03_233356.txt](https://github.com/OpenPHDGuiding/phd2/files/2164166/PHD2_GuideLog_2018-07-03_233356.txt)
[PHD2_GuideLog_2018-07-02_202953.txt](https://github.com/OpenPHDGuiding/phd2/files/2164167/PHD2_GuideLog_2018-07-02_202953.txt)

### DIFF
--- a/calstep_dialog.cpp
+++ b/calstep_dialog.cpp
@@ -46,6 +46,8 @@
 #define MAX_STEPS 60.0
 #define MIN_DECLINATION -60.0
 #define MAX_DECLINATION 60.0
+#define MIN_DISTANCE 1.0 // arc-sec
+#define MAX_DISTANCE 1800.0 // arc-sec (0.5 deg)
 
 static wxSpinCtrlDouble *NewSpinner(wxWindow *parent, int width, double val, double minval, double maxval, double inc)
 {
@@ -71,6 +73,7 @@ CalstepDialog::CalstepDialog(wxWindow *parent, int focalLength, double pixelSize
     m_fPixelSize = pixelSize;
     m_binning = binning;
     m_fGuideSpeed = (float) pConfig->Profile.GetDouble ("/CalStepCalc/GuideSpeed", Scope::DEFAULT_MOUNT_GUIDE_SPEED);
+	m_dDistance = pConfig->Profile.GetDouble("/scope/CalibrationDistance", DEFAULT_DISTANCE);
 
     // Now improve on Dec and guide speed if mount/pointing info is available
     if (pPointingSource)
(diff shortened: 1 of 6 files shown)


alexhelms:
Surprisingly at my image scale I was having very good guiding performance. On par and sometimes better than with my 127mm APO with the same guide camera. The guide camera is a QHY5III178 which has 2.4um pixels and I am already binning it 2x2 (it can't go any higher since CMOS). With CMOS becoming more popular and with pixel size shrinking, I think having the option to override the calibration distance makes sense for some users. A default value will suffice for probably almost everybody. Also, a longer distances gets more points for the line fit for determining the direction of the axes.

I'll make those changes and push them up shortly.

I'll also make sure the distance is computed in the wizard to a default value. That is a good idea and something I definitely overlooked.

Lastly I'm not very familiar with the AO capabilities and features in PHD2 so I'm not sure if I can answer that.

agalasso:
:+1:

PR #708, opened 2018-07-17
Z-filter frequency domain filter

ken-self:
New algorithm for RA and Dec axes with a selection of options. Also an option to show corrections on the same scale as the raw distances.


zfilterfactory.h
```diff
@@ -0,0 +1,96 @@
...
+class ZFilterFactory
+{
+public:
+    std::vector<double> xcoeffs, ycoeffs;
+    double gain() { return ::hypot(dc_gain.imag(), dc_gain.real()); };
+    double corner() { return 1 / raw_alpha1; };
+    int order() { return m_order;  };
+    ZFilterFactory( FILTER_DESIGN f, int o, double p );
+private:
+    const double  PI = 3.14159265358979323846;  // Microsoft C++ does not define M_PI !
```
agalasso: M_PI is defined in <math.h> in MS C++
can remove this definition and include <math.h>

ken-self:
Thanks Andy. Made the suggested changes.

agalasso:
:+1:

PR #732, opened 2018-11-09
Extend DEB package build dependency of libnova

tstibor:
Most current and upcoming DEB package based distributions
such as Ubuntu and Debian ship libnova in version 16.
Extend the package build dependency to encompass also libnova-0.16-0.


agalasso:
thanks!

PR #744, opened 2019-01-19
Touptek on Linux

ken-self:
I took the easy way out and changed S_OK to 0 on line 854 to get around a compiler error. You may want to define S_OK conditionally for Linux.
I changed from trigger mode to video mode as in trigger mode on both Win and Linux a 1s exposure led to a 1.6s sampling period. In video mode a 1s exposure is very close to a 1 sampling period.
In early testing of video mode I was getting an occasional long delay between frames but I have not been able to reproduce it recently. Could have been a USB issue.

While testing with a very artificial "star" I saw fairly large (0.5 pixel), high frequency deviation on the dec axis. What was weird is that when I turned on the target plot the mean position of the deviations switched to the other side of the axis. Ans when I switched it off again the mean returned to its previous location.  I'll post details to the forum as it may be unrelated to the toupcam. 

toupcam.h was saved to phd2/cameras/toupcam.h
libtoupcam.so was saved to /usr/lib/x86_64-linux-gnu/libtoupcam,so

Tested on Ubuntu 16.04 64bit VM

### DIFF
--- a/cam_touptek.cpp
+++ b/cam_touptek.cpp
@@ -563,7 +563,7 @@ bool CameraToupTek::Connect(const wxString& camIdArg)
     m_cam.SetOption(TOUPCAM_OPTION_DFC, 0);
     m_cam.SetOption(TOUPCAM_OPTION_SHARPENING, 0);
     m_cam.SetOption(TOUPCAM_OPTION_AGAIN, 0);
-    
+
     unsigned short speed;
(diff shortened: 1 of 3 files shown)


agalasso:
ok, I will contact ToupTek.  Could you tell me which camera model you are using so I can tell them?

If you want to modify the PR to remove the video mode change I can merge it.

ken-self:
Mine is a GCMOS01200KMB.
Sorry about the force push - got some commits mixed at my end.

PR #753, opened 2019-02-19
Adding INDI SBIG AO driver support

knro:
This adds support for INDI SBIG AO driver. It was tested with SBIG AO-8


agalasso:
awesome, thanks!

Issue #773, opened 2019-06-25
Allow to resize polardrift window for better reading status bar messages


tstibor:
Polardrift window cannot be resized (OS: Debian 10, XFCE4 WM),
see [screenshot1](http://stibor.net/phd2/Screenshot_2019-06-25_11-10-16.png), [screenshot2](http://stibor.net/phd2/Screenshot_2019-06-25_11-10-46.png), [screenshot3](http://stibor.net/phd2/Screenshot_2019-06-25_11-11-05.png) and as a result the
statusbar message cannot be read completely

tstibor:
Closed via https://github.com/OpenPHDGuiding/phd2/pull/774

PR #774, opened 2019-06-25
Allow to resize polardrift window for better reading status bar msg

tstibor:
This fixes issue https://github.com/OpenPHDGuiding/phd2/issues/773


agalasso:
excellent, thanks for the PR.   It can be merged, but while you are at it, do you want to add some code to save the window size across sessions (maybe the position too?) so the user does not have to resize it every time? There's some sample code that saves/restores window position in DriftToolWin::OnClose() and DriftToolWin::DriftToolWin()

tstibor:
Yes, that's a good idea. I will update the PR with this feature.

Issue #770, opened 2019-06-19
request to have a button for auto-star selection


agalasso:
1. a button would be more up-front in the UI (we encourage users to use star auto-selection)
2. button is easier then keyboard shortcut when wearing gloves for example, and one fewer click than accessing it via the tools menu

agalasso:
fixed by 0ef16fabdbf48c31c606a0519ad8d95f7f37fe04

PR #793, opened 2019-08-28
Added a button for auto selecting a star #770

PawelPleskaczynski:
In this PR I added a button for auto selecting a star as proposed in #770


agalasso:
Hi Pawel,
Thank you for the PR! It looks very clean and straightforward.
Bruce and I (maintainers) have been discussing having PHD2 automatically select the guide star when the Guide button is pressed which would make this PR unnecessary, but we have not made a decision on that yet.  I'm going to create a new feature request to capture the discussion Bruce and I had about auto star selection; that way you can be in the loop on where we go with that.

PawelPleskaczynski:
Hello Andy,
Okay, I'll wait 😃

bwdev01:
I have a few comments on the PR itself independent of what we decide to do
about the one-button-guiding option.  To be consistent with the
left-to-right sequencing of options in the toolbar, I think the new button
should be placed to the left of the guiding button.  Second, it should be
disabled when guiding is active, same as the equivalent  pulldown menu item.
Other than that, it looks pretty good and I do like the icon - way better
than the crummy-looking stuff I usually come up with. <g>
 
More later on revisiting the idea of the one-button-guiding alternative.
 
Bruce
 
  _____

PawelPleskaczynski:
I forgot about disabling it while there's guiding, I'll change it

Haha thanks, I made it in GIMP in like 5 minutes :stuck_out_tongue:

myframe.cpp
```diff
@@ -1077,6 +1089,7 @@ void MyFrame::UpdateButtonsStatus()
     if (!guiding_active ^ m_autoSelectStarMenuItem->IsEnabled())
     {
         m_autoSelectStarMenuItem->Enable(!guiding_active);
+        cond_update_tool(MainToolbar, BUTTON_AUTOSTAR, m_autoSelectStarMenuItem, !guiding_active);
```
agalasso: so as to only trigger the ui refresh when the button state changes should be:
```
if (cond_update_tool(...))
    need_update = true;
```

agalasso: please disregard, I see you already have the chage detection in the outer conditional

bwdev01:
It looks ok to me.  The 'Main Screen' and 'Basic Use' sections of the Help guide will need to be changed.

PR #794, opened 2019-08-29
Star is selected automatically after clicking guide button

PawelPleskaczynski:
This PR is alternative to #793 (and personally I think it's better after testing). The star is selected automatically only if user didn't select his own star manually


bwdev01:
Hi Pawel.  I don't think this is a good idea at all.  Auto-select should be
a declarative action, it should always do its thing - the same as with the
menu item or the keyboard shortcut.  Users can have good reasons for wanting
to do it even if another start has been selected.  They may have chosen a
bad star manually (e.g. the brightest), they might have accidentally clicked
on the display, whatever.  
 
FWIW, this is why Andy and I usually spend a fair amount of time discussing
UI changes before trying to implement them.  It's important to get it right.
:-)
 
  _____

agalasso:
yeah, this is the kind of thing Bruce and I had discussed.
The concern we had is that PHD2 will have two slightly different behaviors depending on whether a star is selected or not when the user clicks the Guide button:
 - if a star is already selected, then guiding starts on the selected star (like the current behavior)
 - if a star is not selected, then phd2 automatically selects one (this is the desired behavior in most cases)

The implicit auto-selection is the problem --- it probably won't be obvious to users what is happening or why.  Most of the time we want PHD2 to auto-select a star, but we also want users to have the ability to override that behavior and manually click the star like they do today.

On solution might be to have a mode selection available somehow:
   - auto-selection mode: phd2 always auto-selects the guide star
   - manual selection mode - (just like current behavior) - user selects the guide star either by clicking on it or auto-selecting one with Alt-S or the menu.

PawelPleskaczynski:
Yeah, that are valid points. The change seemed to be better for me because I knew it was there, but users won't know it :sweat_smile:

agalasso:
yeah, I think the challenge is to be able to indicate the two modes -- always-auto-select mode vs manual select mode -- in the UI in a simple way that will not add a lot of confusion ... looking for ideas :confused:

PawelPleskaczynski:
Maybe an entry in one of toolbar's menus? I think toggle "Auto-select guide star" under "Guide" would be good enough

bwdev01:
I think this is getting way too complicated.  I don't think we want to drive
people into the AD to accomplish simple things like getting a guide star
selected.  I suppose we could consider a hybrid implementation.  We could
add the auto-select button, which is what was asked for and will hopefully
be easily understood.  Then we could quietly add the logic to auto-select a
star if none is selected and the user clicks on Guide.  The Guide button is
technically non-deterministic already because it might invoke a calibration.
And some users have been surprised, they don't know why they're getting (or
not getting) the "funny yellow cross-hairs" - but they forge ahead.  In this
scenario, people who have a standard way of doing things will not be
disrupted, they won't even know that anything has changed (other than the
UI).  It's not unlike the keyboard shortcuts that people don't know about.
But over time or with new users, they can discover the new feature and use
it or not.  And using it would be the epitome of P-H-D.
 
  _____

PawelPleskaczynski:
Could you check out my latest commit? In it user can decide whether PHD auto-selects a star on starting guiding or not (defaults to no). If it's checked, you can both let PHD to select a star or click a star manually, otherwise the behavior is the same as in previous PHD versions. What do you think about this?

agalasso:
it's ok, but not great IMO.
what do you think about Bruce's latest proposal:
  - add the auto-select button on the toolbar
  - have the code auto-select a star if no star is already selected

PawelPleskaczynski:
>  the code auto-select a star if no star is already selected

What exactly do you mean by this? The program auto-selects a star if no star is selected after clicking Guide button or the proposed Auto-select button?

agalasso:
sorry, yes, Bruce's proposal was that the Guide button would cause a star to be auto-selected if no star is already selected

PawelPleskaczynski:
Oh okay! So that's exactly what my other PR does

agalasso:
cool, let me give that another look.  If it looks ok and Bruce agrees then we will go ahead and merge that one.

PawelPleskaczynski:
I just wanted to show how PHD behaves after this PR, in my own opinion it's better than the other one (one click for both actions instead of two clicks), please take a look: https://www.youtube.com/watch?v=SzMnbvWtPYE

agalasso:
yes, I agree there is definitely some appeal to this approach and ultimately I'd like to see us get there, but the problem in my opinion is in the UI complexity: we now have two "modes" (as represented by the checkbox on the Guide menu), plus the Auto-select star menu item on the tools menu.  This will be confusing, especially to new users. This is the concern Bruce and I have had and has prevented us from implementing something like this.

PawelPleskaczynski:
Ah, I see! Thanks for the explanation, now I fully agree with you

agalasso:
closing this in favor of PR #793 (merged)

PR #797, opened 2019-09-13
Extend auto-adjustment of guiding params for image scale changes due …

bwdev01:
…to px-size, binning, focal length

### DIFF
--- a/myframe.cpp
+++ b/myframe.cpp
@@ -2742,12 +2742,13 @@ void MyFrame::HandleImageScaleChange(double NewToOldRatio)
 {
     if (pMount)
     {
-        // Adjust the calibration step-size to facilitate a reasonable calibration
-        if (!pMount->IsStepGuider())
+        // Adjust the calibration step-size to facilitate a reasonable calibration of the real mount
+        Scope *scope = TheScope();
+        if (scope)
         {
             int stepSize = ((Scope*)pMount)->GetCalibrationDuration();
             stepSize *= NewToOldRatio;              // Larger scale => larger step-size because calibration is targeted on fixed-size total pixel displacements
-            ((Scope*)pMount)->SetCalibrationDuration(stepSize);
+            scope->SetCalibrationDuration(stepSize);
 
         }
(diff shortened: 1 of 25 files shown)


myframe.cpp
```diff
@@ -2735,27 +2735,38 @@ void MyFrame::RegisterTextCtrl(wxTextCtrl *ctrl)
...
-void MyFrame::HandleBinningChange()
+// Reset the guiding parameters and the various graphical displays when image scale is changed outside the AD UI.  Goal is to restore basic guiding behavior
+// until a fresh calibration is done.  Ratio of image scales is used because 1) info like mount guide speed may not be available and 2) the user may have already 
+// adjusted the calibration step-size in the profile to get the results he wants.
+void MyFrame::HandleImageScaleChange(double NewToOldRatio)
 {
     if (pMount)
     {
+        // Adjust the calibration step-size to facilitate a reasonable calibration
+        if (!pMount->IsStepGuider())
```
agalasso: I think this should be `Scope *scope = TheScope(); if (scope) { int stepSize = scope->GetCalibrationDuration(); ... scope->SetCalibrationDuration(...`
so that it updates the scope calibration step size for the scope when the primary "mount" is the ao.

myframe.cpp
```diff
@@ -2742,12 +2742,13 @@ void MyFrame::HandleImageScaleChange(double NewToOldRatio)
 {
     if (pMount)
```
agalasso: `if (pMount)` can go away too

myframe.cpp
```diff
@@ -2742,12 +2742,13 @@ void MyFrame::HandleImageScaleChange(double NewToOldRatio)
 {
     if (pMount)
     {
-        // Adjust the calibration step-size to facilitate a reasonable calibration
-        if (!pMount->IsStepGuider())
+        // Adjust the calibration step-size to facilitate a reasonable calibration of the real mount
+        Scope *scope = TheScope();
+        if (scope)
         {
             int stepSize = ((Scope*)pMount)->GetCalibrationDuration();
```
agalasso: should be `scope->GetCalibrationDuration()`

PR #801, opened 2019-09-23
WIP Software binning

PawelPleskaczynski:
Hello! I tried to add software binning but since I'm not very familiar with WX framework, I didn't know how to change size of the image in `worker_thread.cpp` or `image_math.cpp` (using setters to change width and height crash the program). What could be solution to this?
Also, if it gets sorted out, what would be the best way to bin images? I thought about using non-overlapping subarrays of fixed size, averaging them and then setting these values as pixel values on the image, would that work?


agalasso:
Hi Pawel,

I think we need to consolidate software binning with the camera hardware binning. Some reasons:

a) changing binning (whether via hardware of software) has ramifications all over the app -- dark libraries and bad-pixel maps are invalidated, pixel scale ("/px) changes, guiding parameters change, calibration step size needs to change, subframe and lock pos invalidated, etc.  PHD2 currently deals pretty well with binning changes, but only via the camera (hardware) Binning property.

b) adding a separate UI control for software binning distinct from hardware binning could potentially be confusing for users, especially if the controls can interact with each other in unexpected ways (e.g. user has hardware binning set to 2, then sets software binning to 2... what do users expect will happen? what will happen?).  Especially confusing since the two controls are in two different areas of the ui.

I think a possible solution to all this would be to extend the behavior of the existing single binning control to provide both hardware and software binning.  Currently the control is populated with the binning values offered by the camera, but we could augment that with additional software binning levels (up to bin 4).  Changing the binning value would have the same effects as it does today (covering all the issues described above in (a)), but if the binning value exceeded the camera's max hardware binning value then phd2 would do the binning in software.

Actually, one of the cameras, cam_touptek.cpp, already does software binning in this way.  The difficulty with this strategy is that each of the cameras (cam_*.cpp) would need to be modified to do something similar to cam_touptek.  There is an opportunity to refactor the camera interface a bit to make it easier to integrate software binning, maybe you can come up with something (in other words, have the software binning be done after the image is returned by Camera::Capture or something like that.)

Some other issues come to mind as I was writing: we'd need to make sure that debayering for color cameras and also dark subtraction interact well with software binning.  Ideally debayering and dark subtraction should be done before software binning, which implies the darks and bad-pixel maps should be stored in un-binned format.  Also, we should probably be smart about combining sw and hw binning, like if the user requests bin4 and the camera can do bin2, we should let the camera do bin2 and then do bin2 in software.

Out of curiosity, do you have a specific camera that motivated you to implement SW binning?

Andy

PawelPleskaczynski:
Hi Andy,
I totally agree with your suggestions, and having two binning settings would be confusing to the user. Well, that's harder than I expected, but I'll try to come up with something, that's a cool exercise!

I wanted to add the software binning because my ASI224MC camera comes only with 2x2 binning and I was curious what would be the effect of guiding with 4x4 and also I wanted to learn some C++ (more than in my school :sweat_smile:).

PR #837, opened 2020-02-09
FreeBSD 12.1: Add compilation support

andkem:
Hello!

This PR contains changes required to make PHD2 compile on FreeBSD. The change highlights are that I've added definition checks for FreeBSD in a few places in the code, worked around a broken find_package for wxWidgets and excluded binary libraries that wouldn't link on FreeBSD.

I'm primarily interested in using Indi features together with PHD2 and I've upstreamed fixes to Indi to make it compile on FreeBSD again. At the moment, the only thing I've been able to properly test is webcam support via Indi using the V4L CCD driver that works on FreeBSD via webcamd. Attaching a webcam and getting images into PHD2 is working with the changes in this PR. Unfortunately, I am currently not able to test using my Celestron mount since it and I are geographically separate.

One question I have is whether the OpenPHD project interested in having upstreamed FreeBSD support? I'm contemplating looking into packaging Indi and PHD2 as a port and it would be smoother if build support were integrated upstream, rather than maintaining a patch set.

To compile on FreeBSD 12.1, do the following:
Make sure the following libraries are installed and available:
cfitsio
eigen3
googletest
libindi (currently not packaged for FreeBSD but should compile using latest upstream)
libusb (provided by the base system)
wx31-gtk3

Configure the project using the following line:
cmake . -DUSE_SYSTEM_CFITSIO=ON -DUSE_SYSTEM_EIGEN3=ON -DUSE_SYSTEM_GTEST=ON -DUSE_SYSTEM_LIBINDI=ON -DUSE_SYSTEM_LIBUSB=ON
make

Thank you for reading!

### DIFF
--- a/contributions/MPI_IS_gaussian_process/CMakeLists.txt
+++ b/contributions/MPI_IS_gaussian_process/CMakeLists.txt
@@ -79,37 +79,43 @@ set_property(TARGET GPGuider PROPERTY FOLDER "Contributions/")
 # Unit tests
 #
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+    set(gtest_link GTest::GTest)
+else()
+    set(gtest_link gtest)
+endif()
+
 
 # Test for the GP
 add_executable(GaussianProcessTest ${gaussian_process_root_dir}/tests/gaussian_process/gaussian_process_test.cpp)
-target_link_libraries(GaussianProcessTest MPIIS_GP gtest)
+target_link_libraries(GaussianProcessTest MPIIS_GP ${gtest_link})
 target_include_directories(GaussianProcessTest  PRIVATE ${gaussian_process_root_dir}/tools ${GTEST_HEADERS})
 set_property(TARGET GaussianProcessTest PROPERTY FOLDER "Unit tests/Contribution")
 add_test(NAME GaussianProcessTest COMMAND GaussianProcessTest)
 
 # Test for the math tools
 add_executable(MathToolboxTest ${gaussian_process_root_dir}/tests/gaussian_process/math_tools_test.cpp)
-target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS gtest)
+target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS ${gtest_link})
 target_include_directories(MathToolboxTest  PRIVATE ${gaussian_process_root_dir}/src ${GTEST_HEADERS} ${EIGEN_SRC})
 set_property(TARGET MathToolboxTest PROPERTY FOLDER "Unit tests/Contribution")
 add_test(NAME MathToolboxTest COMMAND MathToolboxTest)
 
 # Test for the GP Guider
 add_executable(GPGuiderTest ${gaussian_process_root_dir}/tests/gaussian_process/gp_guider_test.cpp)
-target_link_libraries(GPGuiderTest MPIIS_GP gtest GPGuider)
+target_link_libraries(GPGuiderTest MPIIS_GP ${gtest_link} GPGuider)
 target_include_directories(GPGuiderTest  PRIVATE ${gaussian_process_root_dir}/tools ${GTEST_HEADERS} ${phd_src_dir})
 set_property(TARGET GPGuiderTest PROPERTY FOLDER "Unit tests/Contribution")
 add_test(NAME GPGuiderTest COMMAND GPGuiderTest WORKING_DIRECTORY ${gaussian_process_root_dir}/tests/gaussian_process/)
 
 # Performance Test for the GP Guider
 add_executable(GuidePerformanceTest ${gaussian_process_root_dir}/tests/gaussian_process/guide_performance_test.cpp)
-target_link_libraries(GuidePerformanceTest MPIIS_GP gtest GPGuider)
+target_link_libraries(GuidePerformanceTest MPIIS_GP ${gtest_link} GPGuider)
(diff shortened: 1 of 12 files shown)


contributions/MPI_IS_gaussian_process/CMakeLists.txt
```diff
@@ -82,34 +82,54 @@ set_property(TARGET GPGuider PROPERTY FOLDER "Contributions/")
...
 add_test(NAME GaussianProcessTest COMMAND GaussianProcessTest)
 
 # Test for the math tools
 add_executable(MathToolboxTest ${gaussian_process_root_dir}/tests/gaussian_process/math_tools_test.cpp)
-target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS gtest)
+if (${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+    target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS GTest::GTest)
+else()
+    target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS gtest)
+endif()
```
raffienficiaud: Instead of all those `if/else/endif`, you may define once a variable like `gtest_link` as follow
```
if (${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
    set(gtest_link "GTest::GTest")
else()
    set(gtest_link "gtest")
endif()
```

and use it like this
```
target_link_libraries(MathToolboxTest MPIIS_GP_TOOLS ${gtest_link})
```

andkem: Sure, no problem. It's late here so I'll look into it tomorrow.

The "correct" way to do this, in my opinion would be migrating the Linux build to GTest::GTest as well since this is what find_package sets for Google Test. This would require some modifications to the way the bundled Google Test is built so maybe the variable is preferable?

contributions/MPI_IS_gaussian_process/CMakeLists.txt
```diff
@@ -82,34 +82,54 @@ set_property(TARGET GPGuider PROPERTY FOLDER "Contributions/")
 
 # Test for the GP
 add_executable(GaussianProcessTest ${gaussian_process_root_dir}/tests/gaussian_process/gaussian_process_test.cpp)
-target_link_libraries(GaussianProcessTest MPIIS_GP gtest)
+if (${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
```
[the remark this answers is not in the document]
andkem: See my response to raffienficiaud above.

cameras/qhyccd.h
```diff
@@ -20,14 +20,14 @@
 #if defined (_WIN32)
 typedef CCyUSBDevice qhyccd_handle;
 #endif
-#if (defined(__linux__ )&&!defined (__ANDROID__)) ||(defined (__APPLE__)&&defined( __MACH__)) ||(defined(__linux__ )&&defined (__ANDROID__))
+#if (defined(__linux__ )&&!defined (__ANDROID__)) ||(defined (__APPLE__)&&defined( __MACH__)) ||(defined(__linux__ )&&defined (__ANDROID__)) || defined (__FreeBSD__)
```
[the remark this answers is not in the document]
andkem: I didn't realise this was the case. I'll have a look tomorrow at mitigating this issue on a CMake level instead.

contributions/MPI_IS_gaussian_process/CMakeLists.txt
agalasso: I added #839 so we can fix that separately. I'll leave it up to you whether you want to tackle that or not in this PR -- if not then @raffienficiaud's variable suggestion is fine for now.

CMakeLists.txt
```diff
@@ -250,16 +248,31 @@ if(WIN32)
...
 elseif(UNIX AND (NOT APPLE))
   set(cam_SRC
       ${cam_SRC}
       ${phd_src_dir}/cam_qhy5.h
       ${phd_src_dir}/cam_qhy5.cpp
+
+      ${phd_src_dir}/cam_qhy.cpp
+      ${phd_src_dir}/cam_qhy.h
      )
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
```
andkem: I was not sure how to do this in a nice way. Suggestions are welcome if you have something better than this.

contributions/MPI_IS_gaussian_process/CMakeLists.txt
andkem: I opted for the variable solution for now.

cameras/qhyccd.h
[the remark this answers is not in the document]
andkem: I've moved the files in the CMake, but per the comment below I'm not entirely happy with the solution. Suggestions are welcome.

CMakeLists.txt
agalasso: Actually, you can revert his change and update thirdparty.cmake as follows:
Find the if block that says `if(IS_DIRECTORY ${PHD_PROJECT_ROOT_DIR}/cameras/qhyccdlibs/linux/${qhyarch})`
and update it so that the body of that if block is excluded for FreeBSD.
By eliminating `add_definitions(-DHAVE_QHY_CAMERA=1)` the qhy code is excluded from the build (without needing to futz around with the list of camera source files here.)

andkem:
@agalasso You are completely correct. I must have added the define before I modified the CMake. I had already put the section with the define inside of an if statement.

PR #844, opened 2020-02-26
QHY: fix incorrectly-reported bits-per-pixel value for 16-bit cameras

agalasso:
qhy cameras always appear to be 8 bits per pixel.  Use the value returned by GetQHYCCDChipInfo().


bwdev01:
This looks good but I will test it a bit to make sure the AD will correctly handle the late-binding for the Max-ADU control.  I don't expect to find problems.

PR #856, opened 2020-05-14
Renamed qhylibrary for armv6 to be discoverable by cmake

gionkunz:
Hi

I wanted to build phd2 on my Raspberry Zero (armv6z) and was running into the error that the QHY libraries could not be found. I found that the `cmake` FIND_LIBRARY is looking for qhyccd and the library in `phd2/cameras/qhyccdlibs/linux/armv6` was named `libqhy.a` which CMAKE was not able to discover.


agalasso:
hi @gionkunz thanks for the PR, sorry for the delay getting back to you.
The file in the source tree is a very old version if libqhy.a.  QHY tend to change their SDK packaging and library naming relatively frequently and at some point quite a while ago they changed the name in the SDK from libqhy.a to libqhyccd.a.
If renaming the file works for you, I'm ok with taking the PR, but I wanted to caution you that the file itself is probably very old and out of date.
The alternative would be to update to a newer sdk.
I'll merge this pr now, but if you run into any issues with the old sdk library, we can try the newer QHY SDK.

PR #862, opened 2020-06-24
Polar drift tool: add option to handle mirrored image

ken-self:
This adds an option to Polar Drft Alignment to compensate for a mirrord image e.g. by OAG or diagonal. With a mirrored image the corrections currrently shown are in the opposite direction to what they should be


polardrift_toolwin.cpp
```diff
@@ -250,6 +260,17 @@ void PolarDriftToolWin::OnHemi(wxCommandEvent& evt)
     FillPanel();
 }
 
+void PolarDriftToolWin::OnMirror(wxCommandEvent& evt)
+{
+    int i_mirror = m_mirrorCheck->IsChecked()? -1: 1;
+    pConfig->Profile.SetInt("/PolarDriftTool/Mirror", i_mirror);
+    if (i_mirror != m_mirror)
```
agalasso, line 267: this is fine, but could be simply
```suggestion
    m_mirror =  i_mirror;
```

polardrift_toolwin.cpp
```diff
@@ -250,6 +260,17 @@ void PolarDriftToolWin::OnHemi(wxCommandEvent& evt)
...
+    {
+        m_mirror =  i_mirror;
+    }
```
agalasso, line 270: this is fine, but could simply be
```suggestion
    m_mirror = i_mirror;
```

polardrift_toolwin.cpp
```diff
@@ -190,6 +193,13 @@ wxCAPTION | wxCLOSE_BOX | wxMINIMIZE_BOX | wxSYSTEM_MENU | wxTAB_TRAVERSAL | wxF
     m_hemiChoice->SetToolTip(_("Select your hemisphere"));
     gbSizer->Add(m_hemiChoice, wxGBPosition(gridRow, 1), wxGBSpan(1, 1), wxALL, 5);
 
+    // Next row of grid
+    gridRow++;
+    m_mirrorCheck = new wxCheckBox(this, ID_MIRROR, _("Mirror image"));
+    gbSizer->Add(m_mirrorCheck, wxGBPosition(gridRow, 0), wxGBSpan(1, 1), wxALL | wxALIGN_BOTTOM, 5);
+    m_mirrorCheck->SetValue(m_mirror==-1?true:false);
```
agalasso, line 200: this is fine, but equivalent to
```suggestion
    m_mirrorCheck->SetValue(m_mirror == -1);
```

PR #866, opened 2020-07-05
Revise dark-subtraction algorithm to fix problem with star auto-selection due to invalid pedestal value

bwdev01:
The current dark subtraction algorithm works adequately when the light frames are are lighter than the dark frames, which is typically the case.  But when dark frames are lighter then the lights, the existing Pedestal calculation does not behave well since it is based on a the minimum per-pixel difference.  Bruce did some testing with user data frames and found that basing the pedestal value on the difference between the overall median ADU works much better.
In order to support this, we now calculate the median ADU value for each light frame, and for dark frames when they are loaded. When subframes are in use, the median is computed over the subframe ROI for both the dark and light frames.

### DIFF
--- a/darks_dialog.cpp
+++ b/darks_dialog.cpp
@@ -511,8 +511,11 @@ bool DarksDialog::CreateMasterDarkFrame(usImage& darkFrame, int expTime, int fra
         wxYield();
 
         darkFrame.CalcStats();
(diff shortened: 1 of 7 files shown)


agalasso:
thanks Bruce, I'll take a look later tonight

PR #868, opened 2020-07-08
deselect star when scope starts slewing while looping exposures

### LINKED ISSUE #867: automatically de-select star when looping exposures if mount slews
INDI/ASCOM mounts could detect slewing and automatically de-select the guide start.  not much point in trying to keep the star selected when the mount is slewing.

agalasso:
While looping exposures (but not guiding) poll pPointingSource to check for slewing, and deselect the guide star if slewing is detected.
The behavior is enabled only when the _Stop guiding when mount slews_ option is enabled.

Fixes #867

### DIFF
--- a/gear_dialog.cpp
+++ b/gear_dialog.cpp
@@ -1308,6 +1308,9 @@ void GearDialog::OnChoiceScope(wxCommandEvent& event)
             throw THROW_INFO("OnChoiceScope: m_pScope == NULL");
         }
 
+        m_pScope->EnableStopGuidingWhenSlewing(pConfig->Profile.GetBoolean("/scope/StopGuidingWhenSlewing",
+            m_pScope->CanCheckSlewing()));
+
         m_ascomScopeSelected = choice.Contains("ASCOM");
     }
     catch (const wxString& Msg)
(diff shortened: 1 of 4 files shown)


agalasso:
> Concerned about unwanted behavior for scientific uses

agreed. I was hoping tying the new behavior to the existing config option "Stop guiding when scope slews" would reduce the risk of unwanted side effects?

Ok, here's a scenario where this might be a problem: somebody who uses their mount's hand control set to low speed (centering speed) to nudge the mount to either take up dec backlash or to position a guide star for spectroscopy etc.   In those cases de-selecting would be undesirable.

So I'm convinced now we do not want to take this PR as-is.

Another option would be to check the scope coordinates and only deselect if the coordinates differ from the starting coordinates by some minimum amount (like perhaps the size of the frame).   The problem there is that a Sync operation can cause the coordinates to change by an arbitrary amount even though the scope did not move.

Maybe a combination of the two: mount slewed _and_ coordinates changed by an amount larger then the size of the full frame?

agalasso:
does not seem worth pursuing at this time

PR #893, opened 2020-12-21
Fix crash when wdm camera does not report AvgTimePerFrame

mgreter:
I had an issue with a [TSSSI][1] astro webcam (I know, probably extremely poor for guiding, but just starting 😄).
Since I already owned this camera (looks like a 1:1 Omegon Camera CCD clone) I wanted to give it a try.
So debugged down the issue to a simple zero-division, which was easily fixable.
Also seems this value is only used in the select drop-down, so seems rather safe.
Although it will now report the video modes as "0 fps", still better than crashing 🚀 

image

Thanks! Looking forward to actually try my new auto-guiding mount.

[1]: https://www.teleskop-express.de/shop/product_info.php/info/p1778_TS-Optics-Astro-CCD-Kamera---Mond---Planeten---1-25--Anschluss.html


agalasso (approved):
@mgreter thanks for the pr! :+1:

PR #895, opened 2020-12-31
minor cleanups anround new stats code

agalasso:
- add `const` to const methods
- use wxLongLong_t rather then wxLongLong
- use size_t for variables containing container sizes rather than int eliminating implicit cast


bwdev01 (commented):
I can't tell what changes you actually made - the left/right comparison looks like the entire file.  If this is just C++ house-keeping I'm all for it but if the semantics have changed I'd like to discuss.

agalasso:
> the left/right comparison looks like the entire file.

I inadvertently changed the line-endings in that second commit (677dadb7e6e88275d2d1eb00ebf7b37ee57b9a9c) but I fixed that in the 3d commit (1382ae42b3877f005a263bad4e160ce5074cb5b2).  If you click the Files changed tab you should see just the actual changes (all three commits) -- make sure it says "Changes from all commits" at the top left. Also, If you click the gear icon at the top left, you can select "Hide whitespace changes".

bwdev01 (approved):

PR #933, opened 2021-07-11
Do not force c++ standard on linux

### LINKED ISSUE #932: Request to not force c++ standard on linux
Currently, phd2 forces c++ standard to be at max c++11:
https://github.com/OpenPHDGuiding/phd2/blob/771918f27b11ce5513b7a46def4ae70e5d6f48fc/cmake_modules/compiler_options.cmake#L61-L84

C++14 has been the [default standard](https://gcc.gnu.org/projects/cxx-status.html) since GCC 6 and with GCC 11 the standard is now c++17.

While maintaining the package in Fedora repositories, we prefer not to force a specific standard (I was required to do so at the time I submitted the package to Fedora repositories), so we [simply use](https://src.fedoraproject.org/rpms/phd2/blob/rawhide/f/phd2_2.9.10_std_cflags.patch) whatever default is preferred by GCC.
Apart from some warnings, I don't have any compilation error by doing so. Here are the build logs with [GCC 10](https://kojipkgs.fedoraproject.org//packages/phd2/2.6.10/1.fc33/data/logs/x86_64/build.log) (c++14) and [GCC 11](https://kojipkgs.fedoraproject.org//packages/phd2/2.6.10/1.fc35/data/logs/x86_64/build.log) (c++17).

Do you know of any shortcoming by doing that? Would you consider to drop that constraint?

mattiaverga:
This PR will make PHD2 to use gcc default standard on linux instead of force outdated c++11 or older.

Should not impact Mac builds. Fixes #932


d33psky:
I read the mail thread but i'm still confused about this PR.
Because next we will get requests for the exact opposite PR when things do not compile with pre-C11 compilers.
Can't we set a minimum standard on Linux instead ?

mattiaverga:
@d33psky this doesn't force C11, it just uses gcc default.
Currently PHD2 is using `-std=c++0x` for pre-C11 compilers and `-std=c++11` for everything else. There's nothing in between the two. This PR will not change anything for pre-C11 compilers, but will make newer compilers capable of use their default standards (c++14 from GCC 6.1 up until GCC 10 and c++17 from GCC 11).

I've not found any problem building in Fedora with more recent standards, but there we use system libraries (not the bundled ones) and we're not adding many thirdparty drivers.

Would it be acceptable for you if I modify the PR by adding an option to not use anything above c++11 while using gcc defaults by default?

d33psky:
Ah, I missed that part. The PR is fine with me as-is.

agalasso:
@mattiaverga I think @d33psky brings up a valid point: by keeping the standard to C++11 we ensure that any code added by somebody doing development on Linux won't inadvertently break our Mac or Windows builds which are still using C++11 compilers.
We would eventually catch the problem when we went to do a release and found the Mac build failing, but it would be much better to catch the issue early on.
Do you have any ideas on how we can address this?

mattiaverga:
@agalasso I think the best way to catch these type of problems would be to set up some sort of continuous testing with Travis or similar, which will try to build the project at every PR or commit.
If you prefer to test that manually, I can add the option to force c++11 only on linux also.

If you think this will make your work more complicated without adding much value, I can just retire this PR and continue to apply it only on Fedora builds :-)

PR #947, opened 2021-09-13
Update phd2.desktop

atoy40:
The StartupWMClass is needed to associate the launcher icon (for example a dot under the icon in ubuntu/gnome) to the phd2 window.


agalasso (approved):
@atoy40 thanks for the PR.  The fix looks good to me.

@pchev any concerns?

pchev:
No problem, this is a good change.

PR #956, opened 2021-11-29
use uniform captialization for HFD

agalasso:
existing code uses HFD, newly added code uses Hfd; change new code to use the
prior convention. Also,
  - wrap long lines
  - fix an unused local variable


bwdev01 (approved):
Ok with me.  I was worried the HFD capitalization would make it hard to see what it was but I'm fine with keeping things consistent.  Thanks.

PR #957, opened 2021-12-05
ZWO ASI cameras: update SDK to version v1.20.3

agalasso:


bwdev01 (commented):
Ok with me.  But I'm becoming concerned about this constant churning of the ZWO SDK because it seems to trigger many field problems where users see apparent PHD2 crashes because they aren't religiously updating their camera drivers.   We've had several of these just in the past week.  I sympathize with the users on this, no one wants to update device drivers all the time.  Don't you think it further discourages them from running our latest dev builds?

agalasso:
That is a valid concern about driver & SDK stability. But I do not think an SDK upgrade necessarily requires a USB driver update. A USB driver update would _primarily_ be required when ZWO adds a new camera with a new USB device Product Id (USB PID) and the user wants to start using the new camera model and the old driver does not recognize the new camera's PID. OTOH, there is one version (v3.0.0.11  2020.9.21) in the [USB driver change log](https://astronomy-imaging-camera.com/software/change_log/change_camera_driver.txt) that alludes to bug fixes and not just addition of a new USB PID.
I do agree that when diagnosing a camera issue it would make sense to be on the latest USB driver version, but strictly speaking it should not be necessary to update any time there is an SDK update.

IMO, we should point the users to ZWO support and not hold up the SDK updates -- ultimately this is a problem ZWO should be solving and supporting.  The note I got from ZWO regarding this specific SDK is that it addresses some dew issues on cameras with coolers (I will include that info in the PHD2 ChangeLog).  We are in an awkward position with these SDK updates, but my opinion is that the best thing we can do is keep the SDKs up to date with what the vendors provide -- we're just the middle man in this case (between the user and the camera/drivers/SDK code they purchased).  Users may not understand that, so we're unfortunately stuck having to point it out to them.

d33psky:
I recognize that from the INDI perspective, basically in the same boat as PHD2 here wrt. the ZWO SDK.
I agree with Andy that we should not hold up on the SDK updates.

bwdev01:
Sorry, I wasn’t suggesting that we hold up SDK revisions, I was mostly expressing frustration that it tends to make life difficult for us and our users.  And we’re the app that takes it on the chin with the system-level crash, so all of the complaints come to us – ZWO isn’t going to step in and do anything about that, it’s always our fault.

pchev:
I not look at detail, but maybe some try...except in cam_zwo functions may help to give a useful message for the user and prevent phd2 to crash.

agalasso:
> maybe some try...except in cam_zwo functions may help to give a useful message for the user and prevent phd2 to crash

Unfortunately it is not that simple. In all the crash cases I know about are things that are out of our control, such as an access violation in a worker thread spawned internally by the SDK code.

Actually that does suggest a possibility that we could perhaps intercept fatal exceptions (like access violation) and display a message indicating that the camera SDK has crashed and direct the user to contact the camera vendor. PHD2 would still need to terminate, but we could perhaps at least have the opportunity to direct the user to the camera vendor for support.  Added #958

bwdev01 (commented):
I'm all in favor of any error-trapping we can do but it's not the sort of work I can do.  I don't know how we would monitor non-PHD2 SDK threads and I don't know what sort of error scenarios we could hope to capture in a cross-platform environment.  For example, if the kernel-mode driver fouls up the handling of virtual addresses for returning data to the client, I don't know how we would deal it.  But in the end, doing anything in this area would be better than nothing.

agalasso:
> I don't know how we would monitor non-PHD2 SDK threads and I don't know what sort of error scenarios we could hope to capture in a cross-platform environment.

@bwdev01 those are good questions ... could we continue the discussion of what we can to about SDK error handling to #958 ?

If there are no objections about merging this PR could you please approve it and I will merge it. Thanks.

bwdev01 (approved):

PR #961, opened 2021-12-24
Implement variable exposure delays, useful for very accurate mounts; …

bwdev01:
…apply long delays only for steady-state guiding. #Fixes 725


phd.h
```diff
@@ -79,7 +79,7 @@
 
 #define APPNAME _T("PHD2 Guiding")
 #define PHDVERSION _T("2.6.10")
-#define PHDSUBVER _T("dev3")
+#define PHDSUBVER _T("dev3_VarDelay")
```
agalasso, line 82: need to revert this before we merge
```suggestion
#define PHDSUBVER _T("dev3")
```

agalasso (changes requested):
nice work :+1: 

we'll need to revert that version number change before merging this

myframe.cpp
```diff
@@ -607,6 +607,25 @@ int MyFrame::GetTextWidth(wxControl *pControl, const wxString& string)
...
+            rslt = m_varDelayConfig.longDelay;
+        else
+            rslt = m_varDelayConfig.shortDelay;
+        if (rslt != m_varLastExpDelay)
+            Debug.Write(wxString::Format("Exposure delay set to %d\n", rslt));
+        m_varLastExpDelay = rslt;
+        return rslt;
+    }
+    else
+        return m_timeLapse;
```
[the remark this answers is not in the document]
bwdev01: Good idea.

myframe.h
```diff
@@ -342,6 +356,8 @@ class MyFrame : public wxFrame
     void ResetAutoExposure();
     void AdjustAutoExposure(double curSNR);
     static wxString ExposureDurationLabel(int exposure);
+    const VarDelayCfg& GetVariableDelayInfo() const { return m_varDelayConfig; }
+    void SetVariableDelayInfo(bool varDelayEnabled, int ShortDelayMS, int LongDelayMS);
```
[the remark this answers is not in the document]
bwdev01: Good idea.

phd.h
bwdev01: Yes, for sure.

bwdev01:
These are all good suggestions, thanks for the help.  I've made the code changes and I'll test them again tomorrow before re-committing.

agalasso (approved):

myframe.cpp
```diff
@@ -607,6 +607,23 @@ int MyFrame::GetTextWidth(wxControl *pControl, const wxString& string)
...
+
+    int rslt = 0;
+    if (pGuider->IsGuiding() && PhdController::IsIdle() && !pGuider->IsRecentering() && pMount->GetGuidingEnabled())   // pMount will be valid in any state of guiding
+        rslt = m_varDelayConfig.longDelay;
+    else
+        rslt = m_varDelayConfig.shortDelay;
+    if (rslt != m_varLastExpDelay)
+        Debug.Write(wxString::Format("Exposure delay set to %d\n", rslt));
+    m_varLastExpDelay = rslt;
```
agalasso, line 623: `m_varLastExpDelay` is uninitialized, so it would be better if we explicitly initialize it somewhere.
Also, there is no logging when m_varDelayConfig.enabled changes from true to false.
A possible solution to both of these problems would be to change m_varLastExpDelay from a member variable to a static variable used only in this function. Something like this:
```
// Get either timelapse value or state-dependent variable-exposure-delay, in milliseconds
int MyFrame::GetExposureDelay()
{
    int rslt;

    if (!m_varDelayConfig.enabled)
        rslt = m_timeLapse;
    else if (pGuider->IsGuiding() && PhdController::IsIdle() && !pGuider->IsRecentering() &&
             pMount->GetGuidingEnabled())
    {
        rslt = m_varDelayConfig.longDelay;
    }
    else
        rslt = m_varDelayConfig.shortDelay;

    static int s_lastExposureDelay = -1;
    if (rslt != s_lastExposureDelay)
    {
        Debug.Write(wxString::Format("Exposure delay set to %d\n", rslt));
        s_lastExposureDelay = rslt;
    }

    return rslt;
}
```

mount.h
```diff
@@ -366,7 +366,7 @@ class Mount : public wxMessageBoxProxy
     virtual wxString CalibrationSettingsSummary() const { return wxEmptyString; }
 
     virtual bool CalibrationFlipRequiresDecFlip();
-    virtual bool HasHPDecEncoder() { return false; }
+    virtual bool HasHPEncoders() { return false; }
```
agalasso, line 369: let's make this a const method (indicating that calling it has no side-effects on the state of the object)
```suggestion
    virtual bool HasHPEncoders() const { return false; }
```

scope.cpp
```diff
@@ -495,7 +495,7 @@ bool Scope::CalibrationFlipRequiresDecFlip()
     return m_calibrationFlipRequiresDecFlip;
 }
 
-bool Scope::HasHPDecEncoder()
+bool Scope::HasHPEncoders()
```
agalasso, line 498:
```suggestion
bool Scope::HasHPEncoders() const
```

scope.cpp
```diff
@@ -495,7 +495,7 @@ bool Scope::CalibrationFlipRequiresDecFlip()
     return m_calibrationFlipRequiresDecFlip;
 }
 
-bool Scope::HasHPDecEncoder()
+bool Scope::HasHPEncoders()
 {
     return pConfig->Profile.GetBoolean("/scope/HiResEncoders", false);
```
agalasso, line 500: minor: ideally it would be best to have a member variable containing this attribute -- we can use pConfig as our backing store for settings, but for runtime checks use member variables.  The reason is that going though pConfig to get s simple boolean attribute is a heavyweight mechanism ("[premature pessimization](https://elegantchaos.com/2014/10/27/premature-pessimization-and-the-like.html)") In the cases where we use this (GA) we don't really care, so this is probably ok to leave as-is. The motivation for changing it would be to (a) not perpetuate the pattern of using pConfig in place of member variables for object state, and (b) have an efficient function that could be used if this ever needed to be called on the guiding path (outside GA).

agalasso (commented):
nice work @bwdev01 !

I added some comments for your consideration

scope.cpp
```diff
@@ -110,6 +110,8 @@ Scope::Scope()
     val = pConfig->Profile.GetBoolean(prefix + "/UseDecComp", true);
     EnableDecCompensation(val);
 
+    m_hasHPEncoders = pConfig->Profile.GetBoolean("/scope/HiResEncoders", false);
```
agalasso, line 113: do we need to also set m_hasHPEncoders when the GA is run? it looks like as it is now the GA would update the value in pConfig, but the scope's m_hasHPEncoders would not get updated until the scope object is next instantiated (typically next time the user starts phd2)

bwdev01:
I’m not following you on this one.  The GA doesn’t set the property for HPEncoders, it only interrogates.  It may recommend that LowPass2 be used for Dec but it doesn’t appear to update a profile entry for HasHPEncoders.  The idea is that the presence of encoders can only be gotten from the end-user and that happens in the new-profile-wiz.  Am I confused?

agalasso:
> The GA doesn’t set the property

ah, you're right, sorry, I misread it.  code is good as is!

agalasso (approved):

bwdev01:
No problem, I appreciate the discussions before we merge into the master branch.

PR #965, opened 2022-01-18
Svbony Cameras: update SDK to v1.6.4

agalasso:
Also, increase the timeout grace period from 250ms to 500ms as recommended by Svbony.


bwdev01 (approved):
No problem for me.

PR #968, opened 2022-02-08
Small dev4 changes

bwdev01:


star.cpp
```diff
@@ -123,15 +123,16 @@ static double hfr(std::vector<R2M>& vec, double cx, double cy, double mass)
     return hfr;
 }
 
-bool Star::Find(const usImage *pImg, int searchRegion, int base_x, int base_y, FindMode mode, double minHFD, unsigned short maxADU)
+bool Star::Find(const usImage *pImg, int searchRegion, int base_x, int base_y, FindMode mode, double minHFD, unsigned short maxADU, bool minimalLogging)
```
agalasso, line 126: I like the idea of having optional verbose logging, but bool function params can be pretty opaque and the reader of the calling code won't readily understand the meaning of the true/false being passed.

An equivalent but more readable option would be to create an `enum` either global or in `class Star` so that the callers can pass a named value:
```
enum StarFindLogType {
    FIND_LOGGING_MINIMAL,
    FIND_LOGGING_VERBOSE,
};
```
then callers pass one of those values and the function signature would be:
 ```suggestion
bool Star::Find(const usImage *pImg, int searchRegion, int base_x, int base_y, FindMode mode, double minHFD, unsigned short maxADU, StarFindLogType logging)
```

agalasso (changes requested):
looks good, one small recommendation

bwdev01:
Thanks, good suggestion, I'll make the changes and push it again tomorrow.

  _____

agalasso:
looks good so far. when it is ready for re-review, you can click the circle thingy next to my name in the top-right reviewers list to Re-request review.  That will cause me to get a notification and I'll re-review.
(I assume this is still work in progress since there are some Star::Find callers in guider_multistar.cpp that have not yet been converted to the new signature.)

agalasso (approved):
nice work 👍

Issue #974, opened 2022-03-06
debian build fails on build dependency unavailable


alxgomz:
On Debian 11(.2) the package libwxgtk3.0-dev is not available and building the debian package fails with the error bellow:

```
dpkg-checkbuilddeps: erreur: Unmet build dependencies: libwxgtk3.0-dev 
```

PR #975, opened 2022-03-06
update buildeps for newer package name

alxgomz:
fixes issue #974

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: phd2
 Section: education
 Priority: optional
 Maintainer: Patrick Chevalley <pch@ap-i.net>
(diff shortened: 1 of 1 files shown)


pchev:
This is not enough, libgtk2.0-dev also need to be replaced by libgtk-3-dev.
But it may work because it is already a dependency of libwxgtk3.0-gtk3-dev, in this case it is better to remove any libgtk...

Can you try with:
Build-Depends: debhelper (>= 9), cmake, libwxgtk3.0-dev | libwxgtk3.0-gtk3-dev, libcfitsio-dev, libopencv-dev, libusb-1.0-0-dev, libudev-dev, libv4l-dev, libnova-dev, libcurl4-gnutls-dev

I will also test that on old Debian version without GTK3.

alxgomz:
yes removing libgtk2.0-dev works on debian 11

pchev:
This change work for me, OK to commit.

As we are on this subject, I have two other change to do to the debian files.
- change compat to version 10
- for simplification use the unversioned README.txt  in docs
I make another PR after this one is merged.

TallFurryMan:
`cmake_modules/PHD2Packaging.cmake` also needs patching in order to buid the deb file with cmake.

PR #978, opened 2022-03-18
Fix cmake debian dependencies.

TallFurryMan:
Additional fix to #975 to allow cmake to generate a debian package for Ubuntu 20.04.

### DIFF
--- a/cmake_modules/PHD2Packaging.cmake
+++ b/cmake_modules/PHD2Packaging.cmake
@@ -103,7 +103,7 @@ if(UNIX AND NOT APPLE)
   # set version and arch compatible file name
   set(CPACK_PACKAGE_FILE_NAME "phd2_${CPACK_DEBIAN_PACKAGE_VERSION}_${debarch}")
   # Ubuntu 14.04 compatible minimal dependency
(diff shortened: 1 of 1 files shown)


agalasso:
@pchev do you want to review this before I merge it?

agalasso (approved):

pchev:
I not use the packaging cmake, but for me the change look good.

TallFurryMan:
Tested locally here, I need to report I had an issue with LD_LIBRARY_PATH on `libtoupcam.so`, which resided in `/usr/local/lib` and was not readily available in the observatory. I have not investigated why this specific library, and not others, was causing an issue. There might be a `rpath` issue or something, or just my own system config specifics.

TallFurryMan:
OK this is simply because the lib is in a subdirectory. Running `ldd` against `phd2.bin` returns (subset) :
```
libASICamera2.so => /usr/local/lib/libASICamera2.so (0x00007fbcafcb4000)                  libtoupcam.so => not found                                                                libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fbcafb65000)
```

And:

```
$ find /usr/local -name libtoupcam.so
/usr/local/lib/phd2/libtoupcam.so
```

So that lib doesn't seem to be stored at the right place when packaging with cmake. I can probably provide a PR later today.

pchev:
The missing library path is only when you open phd2.bin directly. The script phd2.sh is installed as /usr/bin/phd2 to solve that.

PR #981, opened 2022-04-26
Add option to ignore Bayer matrix for ZWO color cameras

sebgod:
This is useful when using a 850nm IR pass filter in combination with
the ZWO ASI 462mc camera, as it effectively becomes a monochrome
camera.


agalasso (commented):
Hi @sebgod thanks for the PR!

The PR is making two distinct and independent changes. Let's break this into two separate PRs.
 1. the unit white balance initialization
 2. the option to ignore the Bayer matrix

I have some questions/concerns about the second part (ignore bayer matrix).
 - with an IR pass filter in front of the camera, the matrix is still going to produce a differential response across the RGB pixels, right? If that is the case the I would think we actually _do_ want to continue to average over pixels, right?  We can resolve this question definitively by looking at a raw flat frame with the IR filter in place and looking at the mean (median?) ADU values across the frame for the R, G, and B pixels. Can we do that experiment and upload the flat frame to confirm?
 - if the flat frame shows that ignoring the Bayer matrix is a desirable option, then I think the option is something we would want to make available for all color cameras, not just ZWO, right? If so, we would want to make the changes at a higher level (not down in the ZWO camera)

bvalente22:
@agalasso " I think the option is something we would want to make available for all color cameras, not just ZWO, right"

it depends on the camera. the 462 osc has increased sensitivity in the IR range for RGB. other cameras like the 174mc have more of a bandpass where IR range is suppressed. It was a surprise to me to find out th 462mc was so sensitive in IR, but apparently this idea of using IR filter with osc camera for guiding in ir is making the rounds.

agalasso:
Thanks @bvalente22 . If we have at least one camera where we can demonstrate (with an actual flat frame) the benefit of forgoing the averaging, then we can go ahead and add the option.  The option would be disabled by default so users continue to get the current behavior unless the explicitly opt to change it.  Even better (time permitting!), I
could envision PHD2 automatically detecting whether or not to do the averaging: prompt the user to take a flat frame, then look at the statistics on the R.G.B pixels and if they are uniform, disable averaging.

bvalente22:
@agalasso i have one of all these cameras, i can test if you're interested. I am still not sure the benefit, since IR light is less available than broadband, but maybe i'm missing something here

sebgod:
@agalasso , these two changes are actually needed in combination, since only if all pixels are using non-multiplying white balance (not sure about the correct term here) the output will be uniform.
This is the pixel response w.r.t. to wavelength from ZWO:
image
so after 850nm all pixels have (hopefully) all the same response. Technically this is due to the Bayer matrix using some special material that is transparent to NIR.
I can also provide flat frames (I also wrote my own ASCOM driver for this camera that does essentially the same thing, though I didn't take any flat frames yet).
For actual stars this might take a bit as I'm based in a particularly cloudy part of Australia

sebgod:
Here is a flat frame I took with a EL panel https://1drv.ms/u/s!AlSOWTZ-saVGg7gR59NzAkyWefrHHg?e=hHEnxo 
Not sure I am happy with the results but the 3x3 median filter that is done over each exposure in PHD2 (AFAIK) might take care of that?

agalasso:
> Here is a flat frame I took with a EL panel https://1drv.ms/u/s!AlSOWTZ-saVGg7gR59NzAkyWefrHHg?e=hHEnxo

Thanks for the data. I am now a believer.  I took a 200x200 region from the center of the frame and applied PixInsight's SplitCFA tool to get 4 100x100 images, one per channel (R, G, G, B), then looked at the median ADU values:

|Channel|Median ADU|Relative to G|
| -- | --: | --: |
| R | 19198 | -0.02% |
| G | 19201 | +0.00% |
| G | 19201 | +0.00% |
| B | 19204 | +0.02% |

> Not sure I am happy with the results but the 3x3 median filter that is done over each exposure in PHD2 (AFAIK) might take care of that?

If you are going to end up enabling the 3x3 median filter, than a simpler and nearly equivalent solution would be to just use PHD2's default settings! PHD2 (without this PR) does a 2x2 mean to "de-Bayer" the image (QuickLRecon in image_math.cpp).  That should give you comparable results to disabling QuickLRecon (this PR), then enabling 3x3 median.

agalasso:
> these two changes are actually needed in combination, since only if all pixels are using non-multiplying white balance (not sure about the correct term here) the output will be uniform

@sebgod  PHD2 should be initializing ZWO's white balance regardless of anything else. This will ensure that the ZWO ASI camera is always operating under PHD2 with a known good white balance setting (unit white balance.) This could be important if for example somebody was running some other capture app and tweaked the white balance setting, then connected to PHD2. In that case PHD2 would inherit the non-default WB setting and the user could get poor guiding results.  For that reason we try to initialize camera settings for all cameras when PHD2 connects, and PHD2's current omission of initializing the WB setting for ZWO cameras is a bug.  If you want to send that part of this PR as a separate PR we would be grateful and would merge it right away!

sebgod:
@agalasso , here is that PR which does that part (WB balance on init):
https://github.com/OpenPHDGuiding/phd2/pull/983/files

sebgod:
@agalasso thanks for the explanation. I've rebased this PR on the latest master which already contains the white balance changes.
I've also managed to take two frames with the camera using a cheapish achromatic guide scope with a very sloppy helical focuser, on a tripod without tracking. One second exposure with high gain. ASTAP managed to find the centroids and calculate the HFD so I think PHD2 should be able to do this as well.
Only thing I've done is remove the Bayer pattern from the FITS file to simulate what this change is about.
https://1drv.ms/u/s!AlSOWTZ-saVGg7gVV0be7TEyk78L6Q?e=5F7mLE 
https://1drv.ms/u/s!AlSOWTZ-saVGg7gWLgzfjHJAYtfCMg?e=P3ZKWy

rak306:
I would argue that treating the bayer matrix color camera as mono is optimal for an RGB camera  and should perform better than binning 2x2 the red, blue, and 2 green pixels. When guiding, you don’t know the spectral distribution of the guide star. Each pixel gives information on the centroid, and is useful. So there is no magic about making sure we average different colors.

PR #983, opened 2022-05-01
Set colour balance to unity on connection

sebgod:
This avoids non-deterministic behaviour if white-balance
has been adjusted from another application


agalasso (approved):
perfect, thanks!

PR #1002, opened 2022-08-21
Help rework

bwdev01:
This is a substantial update to the Help system with a lot of additional information that is guiding-related but not strictly tied to application features.  It also includes restructuring and expansion of earlier topics based on forum support experience and accumulated changes to the app.


agalasso (approved):
great stuff, thanks!

PR #1003, opened 2022-08-25
SVB Cameras: make ST4 Pulse guide commands synchronous

PeterABerbee:
SVBPulseGuide returns immediately after being called.    It does not wait for the guide pulse to complete.    A delay is necessary to avoid starting an exposure while the mount is moving.    It is also desirable to avoid calling SVBPulseGuide while a pulse is active.

I have verified this is the designed behavior of SVBPulseGuide with SVBony technical support.


agalasso (approved):
cool, thanks!

@PeterABerbee can I add your name to the contributor list (Help > About) as "Peter Berbee"? You can respond here or contact me at andy.galasso@gmail.com.

PR #1005, opened 2022-08-27
ZWO ASI Cameras: update SDK libs to v1.26

agalasso:
update ZWO ASI camera SDK libs to version 1.26


bwdev01 (commented):
It looks like we're running out of runway with the current function for enumerating camera id's.  Do you have multiple ZWO's for testing the new interface?  Do we know yet if this will screw up the profile info?

agalasso:
> It looks like we're running out of runway with the current function for enumerating camera id's. Do you have multiple ZWO's for testing the new interface? Do we know yet if this will screw up the profile info?

@bwdev01  Are you referring to the deprecation comment for `ASIGetProductIDs`? If so, that won't affect PHD2 since we do not use that API call anyway.

bwdev01:
Ok, great, that's what I was concerned about.

bwdev01 (approved):

PR #1011, opened 2022-10-02
On Linux, install the SVBony SDK libSVBCameraSDK.so to /usr/lib/phd2

### LINKED ISSUE #1010: Error while loading shared library
Hi, on Saturday I updated to the latest version, and after the update I got the following error and the application did not start.

> /usr/bin/phd2.bin: error while loading shared libraries: libSVBCameraSDK.so: cannot open shared object file: No such file or directory

As a workaround I installed the following driver and PHD2 worked again:

> https://www.indilib.org/ccds/svbony-sv305-cameras.html

Please let me know if you need any additional information.

pchev:
This is required because PHD2 is statically linked with this library. 
Installing in  /usr/lib/phd2 make it work the same way as the Toupcam SDK.

$ ldd /usr/bin/phd2.bin
        ...
        libtoupcam.so => /usr/lib/phd2/libtoupcam.so (0x00007fc347a00000)
        libSVBCameraSDK.so => /usr/lib/phd2/libSVBCameraSDK.so (0x00007fc347200000)
       ...

Fix #1010


agalasso (approved):
Thanks, Patrick!

PR #1014, opened 2022-10-25
Bogus guide speeds

bwdev01:
The current ZWO ASCOM mount driver reports mount guide speeds in the wrong units, so the speeds look like 1800 deg/sec.  This breaks the PPEC algorithm and causes PHD2 to crash.


scope.h
```diff
@@ -244,6 +244,9 @@ class Scope : public Mount
     bool IsAssumeOrthogonal() const;
     void HandleSanityCheckDialog();
     void SetCalibrationWarning(CalibrationIssueType etype, bool val);
+    bool ValidGuideRates(double RAGuideRate, double DecGuideRate);
+    bool m_CalDetailsValidated;
+    bool m_bogusGuideRatesFlagged;
```
agalasso, line 249: should we put these up above grouped with the rest of the member variables, like after line 150?

scope_ascom.cpp
```diff
@@ -866,6 +866,18 @@ bool ScopeASCOM::GetGuideRates(double *pRAGuideRate, double *pDecGuideRate)
         }
 
         *pRAGuideRate = vRes.dblVal;
+
+        if (!ValidGuideRates(*pRAGuideRate, *pDecGuideRate))
+        {
+            if (!m_bogusGuideRatesFlagged)
+            {
+                pFrame->Alert(_("Mount is reporting invalid guide speeds, some guiding functions including PPEC will be impaired. Contact vendor for support."),
```
agalasso, line 874: hmm, maybe we should blame the ascom driver more directly in the error message?  I believe there are still some mounts out there where the ascom driver provider is not the mount vendor. Tak is one example, there may be more.
```suggestion
                pFrame->Alert(_("The mount's ASCOM driver is reporting invalid guide speeds. Some guiding functions including PPEC will be impaired. Contact the ASCOM driver provider or mount vendor for support."),
```

scope_indi.cpp
```diff
@@ -1,982 +1,992 @@
...
+            *pRAGuideRate =  gra;
+            *pDecGuideRate = gdec;
+
+            if (ValidGuideRates(*pRAGuideRate, *pDecGuideRate))
+                err = false;
+            else
+            {
+                if (!m_bogusGuideRatesFlagged)
+                {
+                    pFrame->Alert(_("Mount is reporting invalid guide speeds, some guiding functions including PPEC will be impaired. Contact vendor for support."));
```
agalasso, line 804:
```suggestion
                    pFrame->Alert(_("The mount's INDI driver is reporting invalid guide speeds. Some guiding functions including PPEC will be impaired. Contact the INDI driver provider or mount vendor for support."));
```

agalasso (commented):
nice work!

scope.h
bwdev01: Thanks for the quick review.  All these suggestions make sense, I'll go ahead and make the changes.

agalasso (approved):

bwdev01:
Thanks.  Does the “auto merge” function on GitHub work correctly?

PR #1015, opened 2022-11-20
Dev3 changes

bwdev01:
Continue to see situations where mis-behaved mounts can create unhandled exceptions in underlying PPEC libraries - these often cause crashes.  Try/catch is a start on the problem but Windows structured exceptions won't be caught.  Other changes are pretty trivial.

### DIFF
--- a/guide_algorithm_gaussian_process.cpp
+++ b/guide_algorithm_gaussian_process.cpp
@@ -944,8 +944,6 @@ GUIDE_ALGORITHM GuideAlgorithmGaussianProcess::Algorithm() const
 
 double GuideAlgorithmGaussianProcess::result(double input)
 {
-    double control_signal;
-
     if (block_updates_)
         return(0);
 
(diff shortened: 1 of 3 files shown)


guide_algorithm_gaussian_process.cpp
```diff
@@ -952,12 +954,21 @@ double GuideAlgorithmGaussianProcess::result(double input)
         return deduceResult();
     }
 
-    // the third parameter of result() is a floating-point in seconds, while RequestedExposureDuration() returns milliseconds
-    const Star& star = pFrame->pGuider->PrimaryStar();
-    double control_signal = GPG->result(input, star.SNR, (double) pFrame->RequestedExposureDuration() / 1000.0);
-
-    Debug.Write(wxString::Format("PPEC: input: %.2f, control: %.2f, exposure: %d\n",
-        input, control_signal, pFrame->RequestedExposureDuration()));
+    try
```
agalasso, line 957: I don't think we want to take this change for a couple reasons:
 1. the catch block code is unreachable because none of the enclosed functions will ever throw an instance of `wxString`.
 2. it introduces a precedent implying that it is OK to throw an exception out of a method and that callers should be expected to handle those exceptions.

The code base currently uses exceptions only for control flow _within_ a function or method, and we do not allow exceptions to propagate. For error propagation we use good old fashioned C-style error return codes to make the code much easier to reason about--we generally do not need to be concerned with C++ exception handling.

Suggestion: it looks like it may be possible for some of the eigen code to throw a `std::runtime_exception`, so if we want to handle that we could add a catch block (or blocks) inside `GaussianProcessGuider::result`.

Let's address this part as a separate PR since it is unrelated to the rest of the changes in this pr.

agalasso (changes requested):

guide_algorithm_gaussian_process.cpp
bwdev01: Agreed, it was a bad idea.  I've reverted the change in the repo or at least I think I have.

PR #1018, opened 2022-12-11
Add rotator selection to new-profile-wizard

bwdev01:
The higher-end Planewave systems come with rotators and we're seeing an influx of these systems - so we need to handle this in the new-profile-wizard.  I didn't want to waste a bunch of time creating a new icon but someone can do that down the road.

### DIFF
--- a/profile_wizard.cpp
+++ b/profile_wizard.cpp
@@ -50,6 +50,7 @@ public:
         STATE_MOUNT,
         STATE_AUXMOUNT,
         STATE_AO,
+        STATE_ROTATOR,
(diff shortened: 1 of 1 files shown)


agalasso (approved):
@bwdev01 I'm guessing you meant to request my review, not assign this to me, right? I cleared the assignment.

Anyway the code looks good, thanks!

bwdev01:
Yes, thanks Andy, I goofed on ‘review’ vs. ‘assign’.  Sigh.

 

Bruce

PR #1019, opened 2022-12-12
Improve alert messages for camera geometry mismatches

bwdev01:
The ZWO software often scrambles the camera connections and users don't understand the current PHD2 alerts about sensor geometry and dark libraries.


agalasso (approved):
nice improvement!

PR #1034, opened 2023-02-07
Calibration assistant

bwdev01:


myframe_events.cpp
```diff
@@ -1013,26 +1016,60 @@ void MyFrame::GuideButtonClick(bool interactive, const wxString& context)
...
+                    double ha = norm(lst - ra, -12.0, 12.0);
+                    if ((fabs(dec) > 20 && dec < degrees(Scope::DEC_COMP_LIMIT)) || fabs(ha) > 3)
+                        calHere = ConfirmDialog::Confirm(
+                        _("Scope isn't pointing in recommended sky area - run the Calibration Assistant to improve results"),
+                        "/v2_highdec_calibration_ok", _("Calibrate here"), _("Calibration Assistant...")
+                        );
+                    else if (dec > 60)
+                        calHere = ConfirmDialog::Confirm(
+                        _("With the scope pointing this close to the pole, calibration accuracy will be degraded and \n"
+                        "Dec compensation will be ineffective. Calibration within 10 degrees of the pole may fail altogether."),
```
agalasso, line 1047: should we also say "run the Calibration Assistant to improve results" here too? the people who hit this error are the ones who especially need to know!

myframe_events.cpp
```diff
@@ -1013,26 +1016,60 @@ void MyFrame::GuideButtonClick(bool interactive, const wxString& context)
...
+                if (!pPointingSource->GetCoordinates(&ra, &dec, &lst))
+                {
+                    bool calHere = true;
+                    double ha = norm(lst - ra, -12.0, 12.0);
+                    if ((fabs(dec) > 20 && dec < degrees(Scope::DEC_COMP_LIMIT)) || fabs(ha) > 3)
+                        calHere = ConfirmDialog::Confirm(
+                        _("Scope isn't pointing in recommended sky area - run the Calibration Assistant to improve results"),
+                        "/v2_highdec_calibration_ok", _("Calibrate here"), _("Calibration Assistant...")
+                        );
+                    else if (dec > 60)
```
agalasso, line 1044:
```suggestion
                    else if (dec > degrees(Scope::DEC_COMP_LIMIT))
```

myframe_events.cpp
```diff
@@ -1013,26 +1016,60 @@ void MyFrame::GuideButtonClick(bool interactive, const wxString& context)
...
+            if (pPointingSource->PreparePositionInteractive())
+                return;
+
+            if (!TheScope()->IsCalibrated())
             {
-                proceed = ConfirmDialog::Confirm(
-                    _("Calibration this far from the celestial equator will be error-prone.  For best results, calibrate at a declination of -20 to +20."),
-                    "/highdec_calibration_ok", _("Confirm Calibration at Large Declination")
-                    );
+                double ra;
```
agalasso, line 1032: should we move this block to a helper function? Then this code becomes:
```
    bool proceed = CheckForCalibration();
    if (proceed)
        StartGuiding();
```
```
static bool
CheckForCalibration()
{
    if (TheScope()->IsCalibrated())
        return true;
    double ra, dec, lst;
    if (!pPointingSource->GetCoordinates(&ra, &dec, &lst))
    {
        ...
```

phdcontrol.cpp
```diff
@@ -65,6 +65,7 @@ struct ControllerState
     bool saveSticky;
     int autoFindAttemptsRemaining;
     int waitSelectedRemaining;
+    bool dontUseStickyLock;
```
agalasso, line 68: can we invert this to be a positive attribute to reduce double negatives?
```suggestion
    bool useStickyLock;
```

phdcontrol.cpp
```diff
@@ -116,7 +117,7 @@ static wxString ReentrancyError(const char *op)
     return wxString::Format("Cannot initiate %s while %s is in progress", op, ctrl.settleOp == OP_DITHER ? "dither" : "guide");
 }
 
-bool PhdController::Guide(bool recalibrate, const SettleParams& settle, const wxRect& roi, wxString *error)
+bool PhdController::Guide(bool recalibrate, const SettleParams& settle, const wxRect& roi, bool noStickyLocking, wxString *error)
```
agalasso, line 120: let's make this a positive attribute too (and if we converge recalibrate and useStickyLock into an `options` param with option bits, then the option bit could be positive like USE_STICKY_LOCK_POS)
```suggestion
bool PhdController::Guide(bool recalibrate, const SettleParams& settle, const wxRect& roi, bool useStickyLockPos, wxString *error)
```

scope.cpp
```diff
@@ -1001,9 +1001,9 @@ void Scope::CheckCalibrationDuration(int currDuration)
 
     int rslt;
     CalstepDialog::GetCalibrationStepSize(pFrame->GetFocalLength(), pCamera->GetCameraPixelSize(),
-        pCamera->Binning, currSpd, CalstepDialog::DEFAULT_STEPS, 0.0, GetCalibrationDistance(), 0, &rslt);
+        pCamera->Binning, currSpdX, CalstepDialog::DEFAULT_STEPS, 0.0, GetCalibrationDistance(), 0, &rslt);
 
-    if (rslt == currDuration)
+    if (fabs(1.0 - (double)rslt/(double)currDuration) <=  0.20)
```
agalasso, line 1006: since it is not obvious, could we add a comment here explaining what this means?

agalasso (commented):
good stuff Bruce!

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+}
+
+void CalibrationAssistant::ShowError(const wxString& msg, bool fatal)
+{
+    m_pMessage->SetLabelText(msg);
+    if (fatal)
+    {
+        m_pSlewBtn->Enable(false);
+        m_pCalibrateBtn->Enable(false);
+
```
agalasso, line 355:
```suggestion
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+void CalibrationAssistant::GetCustomLocation(int* PrefHA, int* PrefDec, bool* SingleSide, bool* UsingDefaults) const
+{
+    *PrefHA = pConfig->Profile.GetInt ("/scope/CalSlew/TgtHA", defBestOffset);
+    *PrefDec = pConfig->Profile.GetInt("/scope/CalSlew/TgtDec", defBestDec);
+    *SingleSide = pConfig->Profile.GetBoolean("/scope/CalSlew/SingleSide", false);
+    *UsingDefaults = (*PrefDec == defBestDec && *PrefHA == defBestOffset && !*SingleSide);
+}
+
+void CalibrationAssistant::PerformSanityChecks(void)
```
agalasso, line 294: optional: you could eliminate much of the indentation to improve readability by adding some early returns:
```
    if (!pPointingSource || !pPointingSource->IsConnected())
        return;

    double raSpd, decSpd;
    bool error = pPointingSource->GetGuideRates(&raSpd, &decSpd);
    if (error)
        return;

    if (!pPointingSource->ValidGuideRates(raSpd, decSpd))
        return;

    ... and so forth ...

File: calibration_assistant.cpp [hunk 9]
```diff
@@ -0,0 +1,1250 @@
...
+}
+
+void CalibrationAssistant::UpdateCurrentPosition(bool fromTimer)
+{
+    double ra, dec, lst;
+    double hourAngle;
+    const int INVALID_DEC = 100;
+    static double lastDec = INVALID_DEC;
+
+    if (!pPointingSource->GetCoordinates(&ra, &dec, &lst))
```
agalasso, line 419: I'd also decrease indentations here in the same way (I find deep indentation difficult to read since I lose the context of the branches)
```
    bool error = pPointingSource->GetCoordinates(&ra, &dec, &lst);
    if (error)
    {
        // show error message
        return;
    }
    // ... ordinary non-error code follows ...
    ...
    if (m_meridianFlipping)
        return;
    ... ordinary non-meridian flipping stuff ...
    etc
}

File: calibration_assistant.cpp [hunk 10]
```diff
@@ -0,0 +1,1250 @@
...
+    double dec;
+    double eqAltitude;
+    bool pointingEast = false;
+    bool errorSeen = false;
+
+    bestDec = defBestDec;
+    bestOffset = defBestOffset;
+    try
+    {
+        if (pPointingSource && pPointingSource->CanReportPosition())
```
agalasso, line 478:
```suggestion
        if (!pPointingSource || !pPointingSource->CanReportPosition())
            throw ERROR_INFO("CalPositionRecommendations: mount not connected or not reporting pointing info");
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+        {
+            if (pPointingSource->PreparePositionInteractive())
+                return true;
+            if (pPointingSource->GetCoordinates(&ra, &dec, &lst))
+            {
+                throw ERROR_INFO("CalPositionRecommendations: Mount not reporting pointing position");
+            }
+            hourAngle = norm(lst - ra, -12.0, 12.0);
+            if (hourAngle <= 0)
```
agalasso, line 487: C/C++ can do nasty things with implicit conversions so I find it helpful to use explicitly-typed constants rather than letting the compiler do implicit casting. 
In this case that means
```suggestion
            if (hourAngle <= 0.)
```
likewise for 30 => 30. 90 => 90. below    (30. and 30.0 would both be ok -- both are the same double whereas 30 is int)

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+    return errorSeen;
+}
+
+void CalibrationAssistant::ShowExplanationMsg(double dec)
+{
+    wxString slewCond;
+    if (pPointingSource->CanSlew())
+        slewCond = _("Use the 'slew' button to move the scope as close as possible to Dec = 0. ");
+    else
+        slewCond = _("Slew the scope as close as possible to Dec = 0. ");
```
agalasso, line 530: should we maybe tone down "as close as possible" -- I'm imagining people obsessively trying to zero-out their dec unnecessarily.
Maybe something like: "Use the Slew button to move the scope to within 15 degrees or so of Dec = 0"

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+    if (fabs(dec) > 80)
+    {
+        m_pExplanation->SetLabelText(_("Calibration is likely to fail this close to the pole.\n") + slewCond);
+    }
+    else if (fabs(dec) > degrees(Scope::DEC_COMP_LIMIT))
+    {
+        m_pExplanation->SetLabelText(_("Declination compensation will not work if you calibrate within 30 degrees of the pole.\n") + slewCond);
```
agalasso, line 537: I guess this is a pre-existing message but I think this message is confusing for newbs who won't know what dec comp is.  Maybe we should be more direct and just tell them to calibrate closer to dec 0, or to disable dec comp?

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+void CalibrationAssistant::ShowExplanationMsg(double dec)
+{
+    wxString slewCond;
+    if (pPointingSource->CanSlew())
+        slewCond = _("Use the 'slew' button to move the scope as close as possible to Dec = 0. ");
+    else
+        slewCond = _("Slew the scope as close as possible to Dec = 0. ");
+    if (fabs(dec) > 80)
+    {
+        m_pExplanation->SetLabelText(_("Calibration is likely to fail this close to the pole.\n") + slewCond);
```
agalasso, line 533:
```suggestion
        m_pExplanation->SetLabelText(_("Calibration is likely to fail this close to the pole.\n") + slewCond);
```
This comment is for all the translatable messages here: let's not include layout formatting ("\n" or trailing " ") in the message that is being translated -- translators may or may not overlook that when they translate and we want the code to be in control of the layout.
Let's ensure that the translators are presented with sentences they can translate without regard to layout formatting. That way they can focus on the translation proper , and the code can do the layout.

calibration_assistant.cpp
```diff
@@ -0,0 +1,1250 @@
...
+    }
+    else if (fabs(dec) > 20)
+        m_pExplanation->SetLabelText(_("Calibration will be more accurate with the scope pointing closer to celestial equator.\n") + slewCond);
+    else
+        m_pExplanation->SetLabelText(wxEmptyString);
+}
+
+void CalibrationAssistant::InitializeUI(bool forceDefaults)
+{
+    if (pPointingSource && pPointingSource->CanReportPosition())
```
agalasso, line 547: ditto comment about deep indentation vs early returns

agalasso (commented):
thanks for making those changes.   I had a chance to look at calibraton_assistant.cpp and have some mostly cosmetic suggestions there.

eegg.cpp
```diff
@@ -234,6 +235,14 @@ void MyFrame::OnPolarDriftTool(wxCommandEvent& WXUNUSED(evt))
     }
 }
 
+void MyFrame::OnCalibrationAssistant(wxCommandEvent& WXUNUSED(evt))
+{
+    if (!pCalibrationAssistant)
+        pCalibrationAssistant = CalibrationAssistantFactory::MakeCalibrationAssistant();
+    if (pCalibrationAssistant)
+        pCalibrationAssistant->Show();
```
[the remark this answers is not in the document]
bwdev01: This is a poor man's "singleton" solution much like the GA and other non-modal dialogs.  We really only want one instance of the CA active.  The destructor for CA nulls the member variable in Frame so I think the null check works as we want.

calibration_assistant.cpp
bwdev01: This is funny.  I intended it to mean "as close as practical for your site" but as you say, we would have people spending 20 minutes trying to completely zero out the Dec.  We know that many people have significant site visibility problems, some quite severe.  Hopefully the changes will work for both cases.

calibration_assistant.cpp
bwdev01: I've taken a different approach using the 'Wrap' method, much like the profile wizard.  Beyond your points, there's really no way to know how things will appear after translation.  So the long text messages are now wrapped and embedded line feeds have been removed.

eegg.cpp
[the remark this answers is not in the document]
agalasso: @bwdev01  I get the singleton thing, that's all good.
The observation is that by line 242 pCalibrationAssistant is never null, thus we can remove the check for null on line 242.

calibration_assistant.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    wxDefaultPosition, wxSize(474, -1), wxCAPTION | wxCLOSE_BOX)
+{
+    m_Parent = Parent;
+    wxStaticBoxSizer* tgtSizer = new wxStaticBoxSizer(wxVERTICAL, this, _("Target Position"));
+    wxFlexGridSizer* targetPosSizer = new wxFlexGridSizer(1, 5, 5, 15);
+
+    int spinnerWidth = StringWidth(this, "0000");
+    m_pTargetDec = NewSpinnerInt(this, wxSize(spinnerWidth, -1), DefaultDec, -50, 50, 5, _("Target declination for slew, as close to Dec = 0 as possible for your location"));
+    AddTableEntryPair(this, targetPosSizer, _("Declination"), m_pTargetDec);
+    m_pTargetOffset = NewSpinnerInt(this, wxSize(spinnerWidth, -1), abs(DefaultHA), 5, 50, 5, _("Target offset from central meridian, in degrees; east or west based on 'Pointing' buttons"));
```
agalasso, line 1019: should we ask for this value in hours then convert to degrees internally?  probably most planetarium and imaging apps will report hour angle in hours not degrees?
we could even ask for the target RA and compute hour angle ourselves, right? (or maybe not, that's assuming they want to image the target _now_)

bwdev01: A lot of this is for users who don't understand the sky coordinate system and have no concept of why right ascension and declination are in different units.  Someone who is dealing with sky obstructions can probably have an intuitive notion of x degrees from the "center" but they aren't going to translate that into x/15 hours.  I don't think a planetarium app has any meaning here, it's PHD2 that's going to slew the scope.  We're not going to point at a particular target, we're going to point "over there".  In my view, changing the hour angle will be an unusual case and mostly done to deal with obstruction problems (balcony location, trees, neighbor's house...).

calibration_assistant.cpp
```diff
@@ -0,0 +1,1247 @@
...
+
+        if (goodRslt)
+        {
+            ShowStatus(_("Calibration produced a good result!"));
+            Debug.Write("CalAsst: good result\n");
+        }
+        else if (acceptableRslt)
+        {
+            {
+                ShowStatus(_("Calibration produced an acceptable result") + "\n" + evalWhy);
```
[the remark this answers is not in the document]
bwdev01: That was the idea of the 'explain' button.  If they want to know how to get a better grade, they would click on 'explain' and get info on what the specific underlying problem is.  I think the alert levels we have provide the necessary warnings and are important enough that we can defend raising the concern.  What I see on the forum is that any sort of caution or warning causes users to become paralyzed, yell for help or complain on various other forums, perhaps only to be told that they should just forge ahead.  I think dealing in generalities here would in fact open up a can of worms.  "Acceptable" means just that - "it's ok to proceed".

calibration_assistant.cpp
```diff
@@ -0,0 +1,1247 @@
...
+public:
+    CalCustomDialog(CalibrationAssistant* Parent, int DefaultHA, int DefaultDec);
+
+private:
+    CalibrationAssistant *m_Parent;
+    wxSpinCtrl *m_pTargetDec;
+    wxSpinCtrl *m_pTargetOffset;
+    wxRadioButton *m_pTargetWest;
+    wxRadioButton *m_pTargetEast;
+    wxCheckBox *m_pEastWestOnly;
```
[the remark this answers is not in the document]
bwdev01: I'm not following this one.  These variables are members of the CalCustomDialog class and they appear at the top of that class definition.  What am I missing?

calibration_assistant.cpp
```diff
@@ -0,0 +1,1247 @@
...
+            evalWhy += _("Sky location, ");
+        }
+
+        evalWhy = "(" + evalWhy.Truncate(evalWhy.length() - 2) + ")";
+        m_lastResult = evalWhy;
+        Debug.Write(debugVals + "\n");
+
+        if (goodRslt)
+        {
+            ShowStatus(_("Calibration produced a good result!"));
```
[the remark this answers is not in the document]
bwdev01: All I wanted here was a 'thumbs-up' message, didn't want to get into TMI.  People don't like calibration.  Setting auto-restore really depends on whether they set up every night and are being careful to reassemble the guiding assembly the same way each night.  Even the "any sky location" behavior depends on having Dec comp set - admittedly nearly always true.  I can be convinced if you feel strongly about it but sometimes less said is more.

calibration_assistant.cpp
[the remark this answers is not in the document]
agalasso: I was suggesting putting the data members before method declarations, i.e. move lines 130-135 (the data member decls) to after line 125

calibration_assistant.cpp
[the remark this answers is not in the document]
agalasso: agreed, don't want to fall int a TMI trap, which my suggested message probably does.
I am imagining the newbie person running the calibration wiz and getting this message, and then wondering "now what?".   Was trying to suggest we tell them what to do next. Any better ideas?

calibration_assistant.cpp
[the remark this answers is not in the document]
agalasso: hmm, could we actually say "It's ok to proceed."?  that could be helpful! (assuming they know what proceed means -- they might not)

calibration_assistant.cpp
agalasso: ok.    I think what threw me off was "target" -- when I saw that I thought "imaging target" -- but this is really the calibration target location.  Should we avoid the word target to avoid mixing it up with imaging target?  Maybe call it "Calibration position" or "Calibration location" or "Calibration pointing position"?

calibration_assistant.cpp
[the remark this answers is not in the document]
bwdev01: Ok, I see it now - I couldn't see they were below the constructor declaration.

calibration_assistant.cpp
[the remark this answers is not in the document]
bwdev01: At this point, guiding has started.  I suppose I could say guiding has started using the new calibration.  I didn't want to close the dialog because the user might want to iterate or experiment further and I wanted them to see a ta-da message.

calibration_assistant.cpp
bwdev01: Ok, yes, "target" is probably a loaded term.  I will try "Calibration location".  This may also help discourage the notion that you need to choosing a specific star for calibration.

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+
+CalibrationAssistant::~CalibrationAssistant()
+{
+    delete m_pTimer;
+    pFrame->pCalibrationAssistant = NULL;
+}
+
+void CalibrationAssistant::GetCustomLocation(int* PrefHA, int* PrefDec, bool* SingleSide, bool* UsingDefaults) const
+{
+    *PrefHA = pConfig->Profile.GetInt ("/scope/CalSlew/TgtHA", defBestOffset);
```
agalasso, line 288: nit
```suggestion
    *PrefHA = pConfig->Profile.GetInt("/scope/CalSlew/TgtHA", defBestOffset);
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+        int recStepSize;
+        CalstepDialog::GetCalibrationStepSize(pFrame->GetFocalLength(), pCamera->GetCameraPixelSize(), pCamera->Binning, sidRate,
+            CalstepDialog::DEFAULT_STEPS, m_currentDec, recDistance, 0, &recStepSize);
+        if (fabs(1.0 - (double)currStepSize / (double)recStepSize) > 0.3)           // Within 30% is good enough
+        {
+            msg = _("Your current calibration parameters can be adjusted for more accurate results."
+                " Click the 'Recal' button to restore them to the default values.");
+        }
+    }
+    if (msg != wxEmptyString)
```
agalasso, line 341: minor
```suggestion
    if (!msg.empty())
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+            {
+                if (lastDec != INVALID_DEC)
+                    ShowExplanationMsg(dec);
+                lastDec = dec;
+            }
+        }
+        else
+            ShowExplanationMsg(dec);
+    }
+}
```
agalasso, line 459: nit
```suggestion
}

```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+            ShowStatus(_("Click 'Calibrate' to start calibration or 'Cancel' to exit"));
+    }
+    m_pSlewBtn->Enable(true);
+    if (TheScope())
+        m_pCalibrateBtn->Enable(true);
+}
+
+static wxString VectorToString(const wxString& separator, const std::vector<wxString>& vec)
+{
+    if (vec.size() == 0)
```
agalasso, line 785: optional:
```suggestion
    if (vec.empty())
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    m_pEastWestOnly = new wxCheckBox(this, wxID_ANY, wxEmptyString);
+    m_pEastWestOnly->SetToolTip(_("Check this if calibration can only be done on a particular side of the meridian because of nearby obstructions"));
+    if (m_pTargetWest->GetValue())
+        m_pEastWestOnly->SetLabelText(_("Western sky only"));
+    else
+        m_pEastWestOnly->SetLabelText(_("Eastern sky only"));
+
+    wxStaticText* pMessage = new wxStaticText(this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxSize(600, -1), wxALIGN_CENTER_HORIZONTAL);
+    pMessage->SetLabelText(_("If your site location requires a unique sky position for calibration, ") +
+        _("you can specify it here."));
```
agalasso, line 1066: This needs to be a single translatable string, not two separate strings -- the individual pieces may not make any sense translated then concatenated together.
The translator is going to use poedit and get a list of all the strings marked by `_()` and translate them one by one.
The two fragments make sense in English concatenated together, but that does not necessarily mean it will work in other languages. For example, It may be more idiomatic in some other language to phrase it like "specify here the sky position required by your location conditional" but as you have it here it forces the translator to translate the individual pieces ("if your site...", "you can spcify") then concatenates them together which prevents the translator from coming up with the best wording for the full sentence.
```suggestion
    pMessage->SetLabelText(_("If your site location requires a unique sky position for calibration, "
                             "you can specify it here."));
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    if (pPointingSource && pPointingSource->IsConnected())
+    {
+        double raSpd;
+        double decSpd;
+
+        if (!pPointingSource->GetGuideRates(&raSpd, &decSpd))
+        {
+            if (pPointingSource->ValidGuideRates(raSpd, decSpd))
+            {
+
```
agalasso, line 1156: nit
```suggestion
```

calibration_assistant.cpp
```diff
@@ -0,0 +1,1266 @@
...
+CalAssistExplanationDialog::CalAssistExplanationDialog(const wxString& Why)
+    : wxDialog(pFrame, wxID_ANY, _("Explanation of Results"),
+    wxDefaultPosition, wxSize(700, -1), wxCAPTION | wxCLOSE_BOX)
+{
+    const int wrapPoint = 550;
+    const int textHeight = 80;
+    wxBoxSizer* vSizer = new wxBoxSizer(wxVERTICAL);
+    wxStaticText* pHeader = new wxStaticText(this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxSize(680, textHeight), wxALIGN_LEFT);
+    pHeader->SetLabelText(_( "Assuming you have followed all the recommendations of the Calibration Assistant, "
```
agalasso, line 1182: nit
```suggestion
    pHeader->SetLabelText(_("Assuming you have followed all the recommendations of the Calibration Assistant, "
```

myframe_events.cpp
```diff
@@ -1013,26 +1046,36 @@ void MyFrame::GuideButtonClick(bool interactive, const wxString& context)
...
+                if (pPointingSource->PreparePositionInteractive())
+                    return;
+
+                if (UseCalibrationAssistant())
+                {
+                    proceed = false;
+                    if (!pCalibrationAssistant)
+                        pCalibrationAssistant = CalibrationAssistantFactory::MakeCalibrationAssistant();
+                    if (pCalibrationAssistant)
+                        pCalibrationAssistant->Show();
```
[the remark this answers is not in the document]
agalasso: @bwdev01 I think this suggestion is still applicable - the code can be simplified by removing the extraneous check for null on line 1067

agalasso (commented):
thanks for addressing my comments!
Only a couple things I believe we should address before merge
 - the translation issue about the concatenated strings
 - making the tooltips imperative to be consistent with other tooltips in the app

the remaining comments are minor and should not block merging this, but feel free to consider the suggestions

agalasso (approved):
perfect, thanks!

bwdev01:
Thanks for the help.

PR #1035, opened 2023-02-07
Report error if INDI v2 is being used

paolostivanin:
With this change, an error is thrown if a user has INDI 2.0 installed and tries to compile PHD2 using `USE_SYSTEM_LIBINDI`.
When using the bundled INDI, phd2 compiles fine.

We need this check in place until we properly migrate INDI code to the v2 format.

This also partially reverse the previous PR (phdindiclient.h)

### DIFF
--- a/phdindiclient.h
+++ b/phdindiclient.h
@@ -35,7 +35,6 @@
 #define PHDINDICLIENT_H
 
 #include <libindi/baseclient.h>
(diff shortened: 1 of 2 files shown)


pawel-soja:
See https://github.com/OpenPHDGuiding/phd2/issues/1036

paolostivanin:
nice, thanks a lot :smile:

Issue #1038, opened 2023-02-12
build fails with override error


justanenduser:
Fully updated 64 bit  Fedora 37. Fully updated phd2 (pulled from git 02/11/2023). 

[ 48%] Building CXX object CMakeFiles/phd2.dir/scope_indi.cpp.o
In file included from /home/gary/Downloads/phd2/config_indi.h:43,
                 from /home/gary/Downloads/phd2/scope_indi.cpp:41:
/home/gary/Downloads/phd2/phdindiclient.h:49:10: error: ‘bool PhdIndiClient::connectServer()’ marked ‘override’, but does not override
   49 |     bool connectServer() override;
      |          ^~~~~~~~~~~~~

imaluschi:
Same here, Kubuntu 22.10, pulled 02/18/2023

![Screenshot_20230219_095257](https://user-images.githubusercontent.com/15073635/219938382-9f123492-1aef-473d-a690-7529d2cda95c.png)

imaluschi:
Ok, got it. With hint from #1032 use -DUSE_SYSTEM_LIBINDI=ON with cmake.
Then this error is gone
(I'm now running into another problem, " /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libwx_gtk3u_core-3.0.so: undefined reference to `TIFFGetVersion@LIBTIFF_4.0' " and other TIFF's ).

pchev:
I also get the same with the INDI 1.8.7 included with the PHD2 source code.

@pawel-soja  what are the INDI version supported by the connectServer() change?  
It work fine with INDI 2.0.0,  but I see this need many change to the PHD2 thirdparty.cmake.

pawel-soja:
The function has been marked as virtual since version 1.9.9, so I added `override` too quickly.
Thanks @pchev for letting me know about this issue.
I'll prepare a fix soon.

pawel-soja:
For some specific reason, GitHub Actions are not running here?

agalasso:
fixed by #1040

PR #1040, opened 2023-02-19
Build Fix - Use `override` for connectServer since 1.9.9 indi.

pawel-soja:
Fixed compilation error for INDI versions below 1.9.9. Detected bug described in the thread: https://github.com/OpenPHDGuiding/phd2/issues/1038


agalasso (approved):
thanks @pawel-soja !

PR #1043, opened 2023-03-05
have a try

AstroAir:
I modified the basic file directory according to the last idea, but there are many issues to be discussed, so I haven't modified the cmake file

I don't know if my modification is feasible. It's roughly like this: All. cpp and. h files are stored in the src folder; 
```
camera : contains all camera control codes;
gui : stores all the wxwidget files
guide_algorithm : all of the guide algorithm files image : Image processing algorithm
logger : Log generation file
rotator : all of the codes about rotator
scope : scope
stepguider : stepguider
server : Server files
```

The work is far from complete, but I believe it is a good trying.


AstroAir:
There are many files here. I don't know if they are missing test files:
```
savetest.fit
savetest2.fit
saveimage.fit
```

For more library files, my idea is to integrate them into the libs folder:
```
/libs
    /win
    /camera
    /indi
```

I'm not sure if I can change these folders:
```
cam_KWIQGuider
MPI_IS_gaussian_process
```

PR #1048, opened 2023-03-20
Dev5 changes

bwdev01:
The only outstanding changes needing review involve removal of the A-P synchronous pulse-guide alert and the correction to avoid losing secondary stars after a huge dither.

### DIFF
--- a/guider_multistar.cpp
+++ b/guider_multistar.cpp
@@ -754,11 +754,17 @@ bool GuiderMultiStar::RefineOffset(const usImage *pImage, GuiderOffset *pOffset)
                         {
                             m_lockPositionMoved = false;
                             Debug.Write("MultiStar: updating star positions after lock position change\n");
-
                             for (auto pGS = m_guideStars.begin() + 1; pGS != m_guideStars.end();)
                             {
-                                if (pGS->Find(pImage, m_searchRegion, pGS->X, pGS->Y, pFrame->GetStarFindMode(),
-                                    GetMinStarHFD(), pCamera->GetSaturationADU(), Star::FIND_LOGGING_VERBOSE))
+                                PHD_Point expectedLoc = m_primaryStar + pGS->offsetFromPrimary;
+                                bool found;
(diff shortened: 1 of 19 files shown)


star.h
```diff
@@ -104,13 +104,15 @@ class GuideStar : public Star
     unsigned int missCount;
     unsigned int zeroCount;
     unsigned int lostCount;
+    PHD_Point offsetFromPrimary;
```
agalasso, line 107: let's add a comment explaining what this is (offset from primary _what_?)

guider_multistar.cpp
```diff
@@ -754,11 +754,17 @@ bool GuiderMultiStar::RefineOffset(const usImage *pImage, GuiderOffset *pOffset)
                         {
                             m_lockPositionMoved = false;
                             Debug.Write("MultiStar: updating star positions after lock position change\n");
-
                             for (auto pGS = m_guideStars.begin() + 1; pGS != m_guideStars.end();)
                             {
-                                if (pGS->Find(pImage, m_searchRegion, pGS->X, pGS->Y, pFrame->GetStarFindMode(),
-                                    GetMinStarHFD(), pCamera->GetSaturationADU(), Star::FIND_LOGGING_VERBOSE))
+                                PHD_Point expectedLoc = m_primaryStar + pGS->offsetFromPrimary;
```
agalasso, line 759: do we need to store offsetFromPrimary on pGS? could we just calculate the offset right here and use it (without extending Star)?

bwdev01: Well, we're not extending Star, it's an extension to GuideStar.  The offsetFromPrimary is built in AutoFind and records the position of each secondary star relative to the primary star.  During the stabilization process in RefineOffsets, the primary star will have moved a long way (outside of its original tracking region), residing somewhere "between" the original lock point and the new one.  We want to look for the secondary stars where we expect them to be, which is the vector sum of the current primary star location and offsetFromPrimary.  I felt this was the cleanest approach to knowing what those positions are.

agalasso (approved):
nice work, thanks!

PR #1049, opened 2023-03-23
fix compile error in calibration_assistant.cpp

### LINKED ISSUE #1047: Compiling fails in Ubuntu 20.04 and Debian 12
So, I tried to follow the instructions to compile phd2 to my machine. I have Debian 12 installed (Bookworm).
First I installed the dependencies, but `libwxgtk3.0-gtk3-dev` and `wx3.0-i18n` seem not to be part of Bookworm anymore. I tried the next best and installed the recent versions: `libwxgtk3.2-gtk3-dev` and `wx3.2-i18n`.

I cloned the repo, created the tmp directory and ran the cmake command. 
Then I tried to compile (`make`).
The compiling fails. I have uploaded the output here:
https://pastebin.com/d9aBa9RM

As Debian 12 is still the testing branch, I thought I would give Ubuntu 20.04 a try, as suggested in the instructions. I set up a VM with Gnome Boxes, installed the (now correct) dependencies, ran cmake and tried to compile. Again, I got an error. 

Both errors occour at 66%.
You can find the output of the compile process (Ubuntu) here:
https://pastebin.com/4FceErn0

Any suggestions?

agalasso:
somehow the code actually compiles on Windows without this fix !?

Fixes #1047

### DIFF
--- a/calibration_assistant.cpp
+++ b/calibration_assistant.cpp
@@ -806,8 +806,8 @@ void CalibrationAssistant::EvaluateCalibration(void)
     static const double CAL_ALERT_DECRATE_DIFFERENCE = 0.20;                    // Ratio tolerance
     static const double CAL_ALERT_AXISRATES_TOLERANCE = 0.20;                   // Ratio tolerance
     bool ratesMeaningful = false;
-    boolean goodRslt = true;
-    boolean acceptableRslt;
+    bool goodRslt = true;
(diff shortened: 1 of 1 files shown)


bwdev01 (approved):

bwdev01:
Evidently!  Looks like it's picking up a typdef in a MS-only .h file.

PR #1058, opened 2023-04-01
fix compile error on Linux/Mac

agalasso:
fixes #1055

PR #1059, opened 2023-04-03
Fix timing problems that may occur w/ real mounts; assure backlash cl…

bwdev01:
…earing for repeated calibrations


calibration_assistant.cpp
```diff
@@ -36,6 +36,7 @@
 #include "phd.h"
 #include "calibration_assistant.h"
 #include "calstep_dialog.h"
+#include <algorithm>
```
agalasso, line 39: this change was already delivered to master in #1058 -- you can remove this change from the pr's branch (you can make sure your branch has the latest changes from master by either merging or rebasing)
to merge:
```
git fetch
git merge origin/master
```

calibration_assistant.cpp
```diff
@@ -704,6 +705,8 @@ bool CalibrationAssistant::PerformSlew(double ra, double dec)
         else
         {
             m_isSlewing = false;
+            ShowStatus(_("Pausing..."));
+            wxMilliSleep(2000);
```
agalasso, line 709: Seems somewhat arbitrary? Are we waiting for a specific guide state -- if so can we wait for that state explicitly rather than waiting 2s for something?
In any case could you add a comment to the code about why this sleep is needed.

bwdev01: Mount controllers/drivers can report premature slew-completions.  This is normally not a problem for anyone but it can ruin a calibration.  Since the goal of the CA is to maximize the likelihood of a successful result, a couple of additional delays seem like a small price to pay.

calibration_assistant.cpp
```diff
@@ -36,10 +36,11 @@
 #include "phd.h"
 #include "calibration_assistant.h"
 #include "calstep_dialog.h"
+#include <algorithm>
```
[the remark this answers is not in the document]
bwdev01: Git was just being doubly certain...  <lol>

agalasso (approved):

PR #1060, opened 2023-04-04
Replace INDI 1.8.7 by 1.8.3 to fix camera crash, fix #1056

### LINKED ISSUE #1056: PHD2 crash when build with the bundled INDI 1.8.7
When build with the bundled indi-1.8.7.tar.gz , PHD2 crash after receiving the first camera image.
The crash is in indi-1.8.7/base64.c when accessing out[] in from64tobits_fast().

I try to replace INDI 1.8.7 by INDI 1.8.2 provided by Ubuntu 20.04 and it work without problem.

@paolostivanin  , any reason you select 1.8.7?  can it be replaced by 1.8.2?

pchev:
Replace the bundled INDI by version 1.8.3 to fix a crash with the INDI camera.

As mentioned in #1056 we not need to take more time on this crash because it is no more present with INDI 2.0.
But it is best to adapt PHD2 code to compile without deprecated warning before to bundle 2.0.


agalasso (approved):
thanks @pchev !

PR #1066, opened 2023-05-01
fix compile warning about deprecated ptr_fun

agalasso:


bwdev01 (approved):
Seems like quite an opaque solution but it does appear to work...  :-)

PR #1068, opened 2023-05-04
Clarify code in AutoFind, remove unnecessary calculation

bwdev01:


star.cpp
```diff
@@ -688,9 +688,9 @@ static void RemoveItems(std::set<Peak>& stars, const std::set<int>& to_erase)
 
 static bool CloseToReference(const GuideStar& referencePoint, const GuideStar& other)
 {
-    // test whether star is close to the reference star for purposes of detecting duplicates
-    const int minLimitSq = 5 * 5;
-    return other.Distance(referencePoint) < minLimitSq;
+    // test whether star is close to the reference star for purposes of detecting duplicates and improving spacial sampling
+    const int minSeparation = 25;
```
agalasso, line 692: the distance returned by `Distance` is a double, and this constant is getting implicitly converted to a double for the comparison. Let's be explicit:
```suggestion
    const double minSeparation = 25.0;
```

PR #1081, opened 2023-06-13
Change star profile image interpolation to nearest neighbor

alexhelms:
The star profile image is currently bicubic scaled but I'm proposing changing to nearest neighbor which I think makes more sense. Normally you want a high quality scaling like bicubic but specifically for astrophotography we actually like nearest neighbor so we can see the pixels. Not seeing the pixels, in my opinion, is misleading because the star appears to show more detail than there really is.

Below is an example using the camera simulator.

**Bicubic (current):**
image

**Nearest Neighbor (proposed):**
image

### DIFF
--- a/star_profile.cpp
+++ b/star_profile.cpp
@@ -301,7 +301,7 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
         wxBitmap subDBmp = dBmp.GetSubBitmap(wxRect(lkx - sz, lky - sz, sz * 2, sz * 2));
         wxImage subDImg = subDBmp.ConvertToImage();
         // scale by 2
-        wxBitmap zoomedDBmp(subDImg.Rescale(width, width, wxIMAGE_QUALITY_HIGH));
+        wxBitmap zoomedDBmp(subDImg.Rescale(width, width, wxIMAGE_QUALITY_NEAREST));
         wxMemoryDC tmpMdc;
         tmpMdc.SelectObject(zoomedDBmp);
         // blit into profile DC


agalasso (approved):
nice improvement!

bwdev01 (commented):
I have reservations about this if it has only been tested using the simulator.  The simulator produces very nice Gaussian profiles which are not typical of many guiding setups.  I would like to see some testing done on real-world star images that include comatic and astigmatic aberrations before we merge the changes

alexhelms:
Here is a screenshot from a real star from last night. This is from an OAG on a 2100mm focal length scope.

image

Regardless, I don't quite understand why this should have been anything other than nearest neighbor. Any kind of interpolation misleads you into thinking the camera has more resolution than it does.

alexhelms:
And here is another example:

image

bwdev01:
I still don't understand why this is a desirable thing to do. I don't want to be argumentative, but I want to give you my perspective based on 10 years of PHD2 development and end-user support for it. The images you've shown are for badly saturated stars, not stars that should be used for guiding.  What is the effect of this change for difficult guide stars - for example, an 11th magnitude star with an SNR of 20?  It looks like the net effect to the typical user might be to see heavy pixelation and greater brightness variations at the pixel level.  How is this useful to them, what is the benefit?  The Star Profile tool is recommended for reaching critical focus of the guide camera.  How will the pixelated star images help with that?  How will the image renderings behave in periods of poor seeing as you move in and out of focus?  Will the revised rendering help a user to see convergence of focus?  Another common use of the Star Profile tool is to look at HFD values to see if the current Min-HFD setting is appropriate.  If the change in rendering has a strong effect on the HFD values displayed in the Profile window, how will they correlate with currently established Min-HFD values?

PHD2 is used by thousands of people around the world, so we have to manage it as we would a commercial-scale product.  When we are considering changes, the decision is always measured by 1) is this going to make the product easier to use and/or 2) is it going to allow people to get better guiding results or use new/different gear.  I can't answer those questions for this proposed change. When we do make a change that is going to affect everyone but may only benefit some, we always provide a means for people to avoid consequences of the change.  If this change was implemented as-is, it would create a support nightmare.  Users would see what appears to be a seriously degraded star image in the Star Profile window and their reaction is predictable:
1.  Something has gone wrong with my camera
2.  I will change the gain settings, change focus, or change the camera driver.  In the process, equipment profiles will be negatively affected.
3.  I will get on the support forum and ask for help.
4.  While I'm waiting for help, I will de-install PHD2 or revert it to an earlier release.
5.  When the earlier release works, I will post additional messages on the forum announcing that the latest PHD2 release has bugs and should be avoided.

As the "owner" of the change, you would be expected to be actively engaged in dealing with all these support issues, probably not something you want to do.  That suggests to me that the new rendering approach would, at the very least, need to be implemented in the UI as an option whose default value would be use of the current algorithm.  That assumes of course, that we can answer the question of whether this is an important change to make and be able to explain how it will provide practical benefits to the user base. 

Thanks.

cconstantine:
I'm a user and recent contributor (I removed a line of code :) ) of phd2, and the current star rendering is something that has confused me.  It confused me because I thought it was an actual zoom-in of the star, but it was suggesting way more resolution that I have, and the star profile to the left didn't seem to fit any part of the star rendering on the right.  This is because the left *is not* interpolated, but the right *is* interpolated.  I would expect for either both of them to be interpolated or neither.  Interpolating for the profile is, I think, more clearly the wrong answer so not interpolating both would be my expectation.

I like this change.

> I still don't understand why this is a desirable thing to do.

The current display algorithm doesn't show the real data, and I usually like seeing the real data.  I agree with @alexhelms that I don't understand why the current behavior is desirable.  In fact, I thought the current display method was a result of some kind of unfortunate limitation, not a decision that was made.

> The images you've shown are for badly saturated stars, not stars that should be used for guiding.

Not quite.  The first two images (with the star profiles) are from the simulator, and do show over saturated stars, but the following two images (without star profiles) are reported to be from a telescope and do not look saturated.  In fact, it's easier to see that they are or aren't fully saturated with this change.  With the bi-cubic interpolation the fully saturated star previews reads to me as having a rounded top which suggests they aren't fully saturated, and the only way to tell they are fully saturated is by looking at the star profile.

>  What is the effect of this change for difficult guide stars - for example, an 11th magnitude star with an SNR of 20? It looks like the net effect to the typical user might be to see heavy pixelation and greater brightness variations at the pixel level. How is this useful to them, what is the benefit? The Star Profile tool is recommended for reaching critical focus of the guide camera. How will the pixelated star images help with that? How will the image renderings behave in periods of poor seeing as you move in and out of focus?

These are some very good questions.  I would expect this to either have no effect, or a positive effect for all of those situations.  Perhaps we could get some samples of those situations with and without this change?

For the low SNR stars that is mostly a question of the stretch applied to the data before display, and, if anything, doing bicubic interpolation will dampen the already low peak signal.  Our vision is based on variations of brightness, so by having greater brightness variations it'll be easier for us to perceive what is really happening.  I've always been frustrated with the star preview for reaching critical focus because it's all blurry.  I can use it to get the focus close, but for getting the final focus position I have to use the star profile (and computed hfd).  This is because the smaller a feature is the more the bicubic interpolation is lying to me about what the data really looks like.  Seeing the pixels also helps me get a sense for how big the currently selected star is.  Without the pixels it's hard to get a sense of how big the star actually is on the sensor.

>  If the change in rendering has a strong effect on the HFD values...

My understanding is that this change will only effect the rendering of the star to the screen.  It will not change the computed HFD.

I can't really speak to the user-support issues this will (or won't) cause, but if this were made into a checkbox in the settings I would expect most people to go in and check it.

agalasso:
thanks for your input @cconstantine 

> if this were made into a checkbox in the settings

let's not go there!  we've got too many settings already and they end being a support problem in themselves!

@bwdev01 what do you think about trying this change out in your obs to get a feeling for how it looks in action?

bwdev01:
I would be willing to do this if I can find the time.  I’m in the middle of a house move and the monsoon season is looming in NM so I think it will be a close call for me.  But I will make the private build and see if I can try it.  With regard to the checkbox – which I’m convinced is required if we do this at all – I would put it right in the Profile Window, certainly not in Advanced Settings.

alexhelms:
I agree with what @cconstantine has said so no need to repeat.

PHD is the only astro software I can think of that shows the user an interpolated zoomed-in image. Everything else shows the user an image without interpolation (nearest neighbor) so the user can visualize the real data from the sensor. The user isn't seeing "heavy pixelation", they are seeing the raw data from the sensor which is critical to understanding. Interpolation hides the ground truth because it is by definition interpolating data that does not exist.

As @agalasso mentioned, there are already a lot of options and I don't think this warrants another one. I'm not seeing this be the issue you're suggesting. A line item in the changelog to show that the rendering has been corrected to better show the raw data should be enough.

cconstantine:
I managed to get some clear skies last night, even if the seeing wasn't great.  Here's some screenshots of the new render algorithm from my Orion 8" f/3.9 with an oag.

Screenshot_20230616_004150
Screenshot_20230615_214344
Screenshot_20230615_214006

cconstantine:
I've had a few nights of good seeing. Here's a screenshot of what that looks like with my equipment.

After using this build for a few days I've found that I strongly prefer this new rendering mode.

Screenshot_20230621_131251

bwdev01:
I’ve done a switchable version of this that I’ll try out in my observatory sometime in the next few nights.

 

Bruce

jmodrako:
Guys, does it affect a guiding in any way?

alexhelms:
> Guys, does it affect a guiding in any way?

No it does not, it is just the image displayed to the user.

bwdev01:
This will show up in the next dev release.  It’s been implemented as a switchable feature, the user can flip back and forth between the two scaling modes by simply clicking on the image in the profile window.  The default setting, for now, is the same as before but the user’s choice is persisted.

 

Bruce

agalasso:
fixed by #1082 

thanks for the PR and for pointing out the issue!

Issue #1083, opened 2023-06-26
Compilation error after change 7620d8f8dd9a19222e5c07d622b7a0682f61cb65


pchev:
After change 7620d8f8dd9a19222e5c07d622b7a0682f61cb65 PHD2 no more compile on Linux :

/home/pch/source/phd2/phd2/star_profile.cpp: In member function ‘void ProfileWindow::OnPaint(wxPaintEvent&)’:
/home/pch/source/phd2/phd2/star_profile.cpp:323:29: error: cannot bind non-const lvalue reference of type ‘wxBitmap&’ to an rvalue of type ‘wxBitmap’
  323 |         tmpMdc.SelectObject(wxBitmap(subDImg.Rescale(width, width, resizeQuality)));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/wx-3.2/wx/wx.h:53,
                 from /home/pch/source/phd2/phd2/phd.h:47,
                 from /home/pch/source/phd2/phd2/star_profile.cpp:36:
/usr/include/wx-3.2/wx/dcmemory.h:29:33: note:   initializing argument 1 of ‘void wxMemoryDC::SelectObject(wxBitmap&)’
   29 |     void SelectObject(wxBitmap& bmp);
      |                       ~~~~~~~~~~^~~

It compile and work after replacing star_profile.cpp line 323 by:

        wxBitmap wxbmp = wxBitmap(subDImg.Rescale(width, width, resizeQuality));
        tmpMdc.SelectObject(wxbmp);

I am not sure this is the best fix, so not make a PR.

Patrick

agalasso:
fixed by #1084

PR #1084, opened 2023-06-29
Fix compile error on Linux

bwdev01:
Response to compile-time issue noted by Patrick:
https://github.com/OpenPHDGuiding/phd2/issues/1083

### DIFF
--- a/star_profile.cpp
+++ b/star_profile.cpp
@@ -320,7 +320,8 @@ void ProfileWindow::OnPaint(wxPaintEvent& WXUNUSED(evt))
             resizeQuality = wxIMAGE_QUALITY_NEAREST;
             toggleMsg = _("Click image for interpolated view");
         }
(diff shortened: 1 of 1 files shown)


d33psky:
This looks like a compiler bug to me. Glad you found a workaround.

bwdev01:
I agree, this works fine on Windows and I don’t see why it should be a problem.  Patrick was the one who discovered the Linux issue and proposed the solution.  As it turns out, the revision basically mimics the way it was done in earlier releases so I suspect the original author took this approach for the same reason.

agalasso:
it's a legit bug, sorry I did not catch it in the code review of the earlier pr
The wxBitmap instance on line 323 (in the code prior to this pr) gets destroyed after it is selected into tmpMdc -- its lifetime is the scope of the expression on line 323, so after that line the dc has selected a destroyed wxBitmap.

agalasso (approved):

PR #1086, opened 2023-07-13
Add profile-wiz ability to set saturation ADU based on camera bit depth

bwdev01:
Initial implementation for profile-wiz to set saturation ADU based on cam bit depth if known.  I'm still seeing many instances of 16-bit drivers being used with a saturation value of 255.  This doesn't "fix" existing profiles.


agalasso (approved):
nice!

PR #1087, opened 2023-07-20
Group Min/Max HFD controls in UI; clean-up for Max-HFD implementation

bwdev01:
Implementation of Max-HFD to handle clumps of sensor noise, internal reflections, unwanted comet heads and other sources of false star detection. Default value will not change historical behavior of star selection.

### DIFF
--- a/guider.cpp
+++ b/guider.cpp
@@ -218,9 +218,11 @@ void Guider::LoadProfileSettings()
     if (minHFD < GetMinStarHFDFloor())
         minHFD = GetMinStarHFDFloor();
     SetMinStarHFD(minHFD);
+    double maxHFD = pConfig->Profile.GetDouble("/guider/StarMaxHFD", 20.0);
+    SetMaxStarHFD(maxHFD);
 
     double minSNR = pConfig->Profile.GetDouble("/guider/StarMinSNR", 6.0);
-    SetMinStarSNR(minSNR);
+    SetAFMinStarSNR(minSNR);
 
     unsigned int autoSelDownsample = wxMax(0, pConfig->Profile.GetInt("/guider/AutoSelDownsample", 0));
     SetAutoSelDownsample(autoSelDownsample);
@@ -1803,21 +1805,27 @@ EXPOSED_STATE Guider::GetExposedState()
 
     return rval;
 }
-
 void Guider::SetMinStarHFD(double val)
 {
     Debug.Write(wxString::Format("Setting StarMinHFD = %.2f\n", val));
     pConfig->Profile.SetDouble("/guider/StarMinHFD", val);
     m_minStarHFD = val;
 }
-
-void Guider::SetMinStarSNR(double val)
+// Minimum star SNR for auto-find; minimum SNR for normal star tracking is hard-wired to 3
+void Guider::SetAFMinStarSNR(double val)
 {
     Debug.Write(wxString::Format("Setting StarMinSNR = %0.1f\n", val));
     pConfig->Profile.SetDouble("/guider/StarMinSNR", val);
-    m_minStarSNR = val;
+    m_minAFStarSNR = val;
 }
 
+// Maximum star HFD
+void Guider::SetMaxStarHFD(double val)
+{
+    Debug.Write(wxString::Format("Setting MaxHFD = %0.1f\n", val));
+    pConfig->Profile.SetDouble("/guider/StarMaxHFD", val);
+    m_maxAFStarHFD = val;
+}
 void Guider::SetAutoSelDownsample(unsigned int val)
 {
     Debug.Write(wxString::Format("Setting AutoSelDownsample = %u\n", val));
--- a/guider.cpp
+++ b/guider.cpp
@@ -1826,7 +1826,7 @@ void Guider::SetMaxStarHFD(double val)
 {
     Debug.Write(wxString::Format("Setting MaxHFD = %0.1f\n", val));
     pConfig->Profile.SetDouble("/guider/StarMaxHFD", val);
-    m_maxAFStarHFD = val;
+    m_maxStarHFD = val;
 }
 
 void Guider::SetAutoSelDownsample(unsigned int val)
--- a/guider.cpp
+++ b/guider.cpp
@@ -1805,12 +1805,14 @@ EXPOSED_STATE Guider::GetExposedState()
 
     return rval;
 }
+
 void Guider::SetMinStarHFD(double val)
 {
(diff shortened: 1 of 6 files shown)


star.cpp
```diff
@@ -990,14 +998,15 @@ bool GuideStar::AutoFind(const usImage& image, int extraEdgeAllowance, int searc
...
+    double maxHFD = pFrame->pGuider->GetMaxStarHFD();
     foundStars.clear();
     for (std::set<Peak>::reverse_iterator it = stars.rbegin(); it != stars.rend(); ++it)
     {
         GuideStar tmp;
-        tmp.Find(&image, searchRegion, it->x, it->y, FIND_CENTROID, pFrame->pGuider->GetMinStarHFD(), pCamera->GetSaturationADU(), FIND_LOGGING_VERBOSE);
+        tmp.Find(&image, searchRegion, it->x, it->y, FIND_CENTROID, pFrame->pGuider->GetMinStarHFD(), pFrame->pGuider->GetMaxStarHFD(), pCamera->GetSaturationADU(), FIND_LOGGING_VERBOSE);
         // We're repeating the find, so we're vulnerable to hot pixels and creation of unwanted duplicates
-        if (tmp.WasFound() && tmp.SNR >= minSNR)
+        if (tmp.WasFound() && tmp.SNR >= minSNR && tmp.HFD <= maxHFD)
```
agalasso, line 1009: I believe this check is redundant and can be removed since if the HFD > maxHFD then `tmp.WasFound()` will be `false` and the check is unreachable.
```suggestion
        if (tmp.WasFound() && tmp.SNR >= minSNR)
```

star.cpp
```diff
@@ -471,7 +479,7 @@ bool Star::Find(const usImage *pImg, int searchRegion, int base_x, int base_y, F
 
 bool Star::Find(const usImage *pImg, int searchRegion, FindMode mode, double minHFD, unsigned short saturation, StarFindLogType loggingControl)
 {
-    return Find(pImg, searchRegion, X, Y, mode, minHFD, saturation, loggingControl);
+    return Find(pImg, searchRegion, X, Y, mode, minHFD, pFrame->pGuider->GetMaxStarHFD(), saturation, loggingControl);
```
agalasso, line 482: let's keep this method decoupled from Guider -- have the caller pass maxHDF in addition to minHFD.  It will be more consistent that way having them both be passed explicitly rather than one explcit and one implicit.

guider.h
```diff
@@ -422,11 +426,15 @@ inline double Guider::GetMinStarHFDDefault() const
...
+inline double Guider::GetAFMinStarSNR() const
 {
-    return m_minStarSNR;
+    return m_minAFStarSNR;
 }
 
+inline double Guider::GetMaxStarHFD() const
+{
+    return m_maxAFStarHFD;
+}
```
agalasso, line 437: cosmetic, match formatting of surrounding code
```suggestion
}

```

guider.cpp
```diff
@@ -1803,21 +1805,27 @@ EXPOSED_STATE Guider::GetExposedState()
...
+    m_minAFStarSNR = val;
 }
 
+// Maximum star HFD
+void Guider::SetMaxStarHFD(double val)
+{
+    Debug.Write(wxString::Format("Setting MaxHFD = %0.1f\n", val));
+    pConfig->Profile.SetDouble("/guider/StarMaxHFD", val);
+    m_maxAFStarHFD = val;
+}
```
agalasso, line 1828:
```suggestion
}

```

guider.cpp
```diff
@@ -1803,21 +1805,27 @@ EXPOSED_STATE Guider::GetExposedState()
...
-
 void Guider::SetMinStarHFD(double val)
 {
     Debug.Write(wxString::Format("Setting StarMinHFD = %.2f\n", val));
     pConfig->Profile.SetDouble("/guider/StarMinHFD", val);
     m_minStarHFD = val;
 }
-
-void Guider::SetMinStarSNR(double val)
+// Minimum star SNR for auto-find; minimum SNR for normal star tracking is hard-wired to 3
```
agalasso, line 1814:
```suggestion

// Minimum star SNR for auto-find; minimum SNR for normal star tracking is hard-wired to 3
```

guider.cpp
```diff
@@ -1803,21 +1805,27 @@ EXPOSED_STATE Guider::GetExposedState()
 
     return rval;
 }
-
 void Guider::SetMinStarHFD(double val)
```
agalasso, line 1808:
```suggestion

void Guider::SetMinStarHFD(double val)
```

agalasso (commented):
nice work!

bwdev01:
Thanks for the help, I’ll get going on the changes.

guider.h
```diff
@@ -266,12 +267,15 @@ class Guider : public wxWindow
...
     double GetMinStarHFD() const;
     double GetMinStarHFDFloor() const;
     double GetMinStarHFDDefault() const;
-    void SetMinStarSNR(double val);
-    double getMinStarSNR() const;
+    void SetMinStarHFD(double val);
+    double GetMaxStarHFD() const;
+    void SetMaxStarHFD(double val);
+    void SetAFMinStarSNR(double val);
+    double GetAFMinStarSNR() const;
```
agalasso, line 277: curious why these methods have "AF" in the name now, bit the other methods do not have AF?

guider.h
```diff
@@ -165,7 +165,8 @@ class Guider : public wxWindow
     LockPosShiftParams m_lockPosShift;
     bool m_measurementMode;
     double m_minStarHFD;
-    double m_minStarSNR;
+    double m_minAFStarSNR;
+    double m_maxAFStarHFD;
```
agalasso, line 169: could we add some comments describing what these variables are? especially to clarify what AF stands for?  I'm finding it a bit confusing to see minHFD without "AF" and maxHFD with "AF" in the name.
Also might be slightly clearer to put the HFD min and max variables together on adjacent lines since they are logically a pair?

agalasso (commented):
code changes look good.  hoping to get a little more clarity on the names of the new member variables in class Star, thanks

guider.h
bwdev01: The two 'AF' methods only control an SNR floor property used in Auto-Find, it is different from the photometric SNR floor used in 'find' operations.  The mis-named class variable was a left-over from my initial cut at this, which anticipated that a max-hfd property would also be used only in Auto-Find.  Obviously, that didn't work very well.

agalasso (approved):
thanks!

PR #1095, opened 2023-09-11
Fix INDI-related bug to avoid returning lst=0 when it is actually unknown

bwdev01:


bwdev01:
Superceded by a later PR with a better solution.

PR #1098, opened 2023-09-17
Compute LST in PHD2 if INDI driver does not provide it and libnova is not available

agalasso:
This is an alternative to #1095 but provides a computed LST value rather than just returning an error.

The LST computation was already present in scope_manual_pointing.cpp, but has been moved to make it accessible to scope_indi.cpp as well.

### DIFF
--- a/image_math.h
+++ b/image_math.h
@@ -128,4 +128,27 @@ inline static double radians(double degrees)
     return degrees * M_PI / 180.;
 }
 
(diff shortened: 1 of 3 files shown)


bwdev01 (approved):
Much better, I didn't know this was already done in the source base and I didn't want to migrate C# code to do it. :-).  Thanks.

Issue #1106, opened 2023-10-10
ScopeASCOM::SetupDialog will unregister COM scope object even if it is already in-use


bwdev01:
ScopeASCOM always un-registers the COM scope object even if the scope is already connected.  Failure will occur after the Connect Equipment dialog is closed and the first reference is made to any scope property.  Steps to reproduce:
1. Open Connect Equipment dialog, connect to scope
2. Click on ASCOM properties icon
3. Close ASCOM properties dialog
4. Close Connect Equipment dialog

PR #1107, opened 2023-10-10
ScopeASCOM::SetupDialog should not unconditionally destroy COM scope instance

bwdev01:
ScopeASCOM::SetupDialog change to address #1106.

### DIFF
--- a/scope_ascom.cpp
+++ b/scope_ascom.cpp
@@ -136,7 +136,13 @@ wxArrayString ScopeASCOM::EnumAscomScopes()
     return list;
 }
 
-bool ScopeASCOM::Create(DispatchObj& obj, bool *alreadyRegistered)
+static bool AlreadyPresentInGit(GITEntry deviceEntry)
+{
+    IDispatch *idisp = deviceEntry.Get();
+    return idisp != nullptr;
+}
+
+bool ScopeASCOM::Create(DispatchObj& obj)
 {
     try
     {
@@ -145,12 +151,10 @@ bool ScopeASCOM::Create(DispatchObj& obj, bool *alreadyRegistered)
         if (idisp)
         {
             obj.Attach(idisp, NULL);
-            *alreadyRegistered = true;
             return true;
         }
 
         Debug.Write(wxString::Format("Create ASCOM Scope: choice '%s' progid %s\n", m_choice, s_progid[m_choice]));
-        *alreadyRegistered = false;
         wxBasicString progid(s_progid[m_choice]);
 
         if (!obj.Create(progid))
@@ -180,8 +184,8 @@ bool ScopeASCOM::HasSetupDialog() const
 void ScopeASCOM::SetupDialog()
 {
     DispatchObj scope;
-    bool alreadyRegistered;
-    if (Create(scope, &alreadyRegistered))
+    bool prevRegistered = AlreadyPresentInGit(m_gitEntry);
+    if (Create(scope))
     {
         Variant res;
         if (!scope.InvokeMethod(&res, L"SetupDialog"))
@@ -192,11 +196,12 @@ void ScopeASCOM::SetupDialog()
             wxMessageBox(msg, _("Error"), wxOK | wxICON_ERROR);
         }
     }
-    // destroy the COM object now as this reduces the likelhood of getting into a
+    // If we just created the COM object for doing the SetupDialog (device not already connected),
+    // we need to destroy it now as this reduces the likelhood of getting into a
     // state where the user has killed the ASCOM local server instance and PHD2 is
     // holding a reference to the defunct driver instance in the global interface
     // table
-    if (!alreadyRegistered)
+    if (!prevRegistered)
         m_gitEntry.Unregister();
 }
 
@@ -215,8 +220,7 @@ bool ScopeASCOM::Connect()
         }
 
         DispatchObj pScopeDriver;
-        bool alreadyRegistered = false;
-        if (!Create(pScopeDriver, &alreadyRegistered))
+        if (!Create(pScopeDriver))
         {
             wxMessageBox(_T("Could not establish instance of ") + m_choice, _("Error"), wxOK | wxICON_ERROR);
             throw ERROR_INFO("ASCOM Scope: Could not establish ASCOM Scope instance");
--- a/scope_ascom.cpp
+++ b/scope_ascom.cpp
@@ -136,7 +136,7 @@ wxArrayString ScopeASCOM::EnumAscomScopes()
     return list;
 }
 
-bool ScopeASCOM::Create(DispatchObj& obj)
+bool ScopeASCOM::Create(DispatchObj& obj, bool *alreadyRegistered)
 {
     try
     {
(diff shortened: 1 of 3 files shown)


bwdev01:
This is issue 1106, just reported by end-user today.  This involves some bare-knuckle COM code so it needs a careful review.

agalasso:
> This is issue 1106, just reported by end-user today. This involves some bare-knuckle COM code so it needs a careful review.

@bwdev01  small request: could fix the pr title so it is not truncated and add a pr desciprion. 

BTW, when you refer to a github issue or PR and write it as #1106 then github will automatically create a link so the reader does not have to manually look up the issue

scope_ascom.cpp
```diff
@@ -136,7 +136,7 @@ wxArrayString ScopeASCOM::EnumAscomScopes()
     return list;
 }
 
-bool ScopeASCOM::Create(DispatchObj& obj)
+bool ScopeASCOM::Create(DispatchObj& obj, bool *alreadyRegistered)
```
agalasso, line 139: I think this will be a little easier to understand if we change the name and invert the sense of this parameter
```suggestion
// *registered is set to true if the Dispatch object was registered in the
// Global Interface Table
bool ScopeASCOM::Create(DispatchObj& obj, bool *registered)
```
so that if `registered` is returned as `true` then the caller may need to call `Unregister`

I would also suggest setting `*registrered = false` at the top of the function, then set it to `true` right after the call to `m_gitEntry.Register(obj);` as that will ensure that it never gets set to tru prematurely regardless of thrown exceptions etc.

scope_ascom.cpp
```diff
@@ -179,7 +180,8 @@ bool ScopeASCOM::HasSetupDialog() const
 void ScopeASCOM::SetupDialog()
 {
     DispatchObj scope;
-    if (Create(scope))
+    bool alreadyRegistered;
```
agalasso, line 183:
```suggestion
    bool registeredInGit; // whether object just got registered in the global interface table
```

scope_ascom.cpp
```diff
@@ -194,7 +196,8 @@ void ScopeASCOM::SetupDialog()
     // state where the user has killed the ASCOM local server instance and PHD2 is
     // holding a reference to the defunct driver instance in the global interface
     // table
-    m_gitEntry.Unregister();
+    if (!alreadyRegistered)
```
agalasso, line 199:
```suggestion
    if (registeredInGit)

```

scope_ascom.cpp
```diff
@@ -212,8 +215,8 @@ bool ScopeASCOM::Connect()
         }
 
         DispatchObj pScopeDriver;
-
-        if (!Create(pScopeDriver))
+        bool alreadyRegistered = false;
+        if (!Create(pScopeDriver, &alreadyRegistered))
```
agalasso, line 219: if we make the output parameter optional (default value = nullptr) then we can keep this code simple as it was before
```suggestion
        if (!Create(pScopeDriver))
```

scope_ascom.h
```diff
@@ -78,7 +78,7 @@ class ScopeASCOM : public Scope
     wxString m_choice; // name of chosen scope
 
     // private functions
-    bool Create(DispatchObj& obj);
+    bool Create(DispatchObj& obj, bool *alreadyRegistered);
```
agalasso, line 81:
```suggestion
    bool Create(DispatchObj& obj, bool *alreadyRegistered = nullptr);
```

agalasso (commented):
the fix looks good, thanks!

I have some readability/maintainability suggestions.

bwdev01:
I felt like the original fix was rather awkward and a simpler approach could be taken.  I just pushed those changes, let me know what you think.

scope_ascom.cpp
```diff
@@ -136,6 +136,12 @@ wxArrayString ScopeASCOM::EnumAscomScopes()
     return list;
 }
 
+static bool AlreadyPresentInGit(GITEntry deviceEntry)
```
agalasso, line 139: `GITEntry` is not expected to be passed by value. If it is not done properly we can have temporary intermediate values that invoke ~GITEntry and that would be really bad.
Passing by reference would be much better here.  It can be const reference since Get() is a const method.
Also, let's keep the same capitalization convention referring to the GIT in all caps in class names & function names:
```suggestion
static bool AlreadyPresentInGIT(const GITEntry& deviceEntry)
```
The other approach (which I would prefer) would be to make this a member function of the GITEntry class:
```
class GITEntry
{
...
    bool IsRegistered() const
    {
        ...
    }
};
```

scope_ascom.cpp
```diff
@@ -136,6 +136,12 @@ wxArrayString ScopeASCOM::EnumAscomScopes()
     return list;
 }
 
+static bool AlreadyPresentInGit(GITEntry deviceEntry)
+{
+    IDispatch *idisp = deviceEntry.Get();
+    return idisp != nullptr;
```
agalasso, line 142: Get() [adds a reference](https://learn.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-iglobalinterfacetable-getinterfacefromglobal#remarks) to the object so as written this leaks a reference.
The GITObjRef class can be used for this purpose:
```suggestion
    return GITObjRef(deviceEntry).IDisp() != nullptr;
```

scope_ascom.cpp
```diff
@@ -179,6 +184,7 @@ bool ScopeASCOM::HasSetupDialog() const
 void ScopeASCOM::SetupDialog()
 {
     DispatchObj scope;
+    bool prevRegistered = AlreadyPresentInGit(m_gitEntry);
```
agalasso, line 187: let's reverse these two lines to keep the scope of of the variable `scope` to the minimum required (and for for clarity -- it is tightly associated the the Create() call)
```suggestion
    bool prevRegistered = m_gitEntry.IsRegistered();
    DispatchObj scope;
```

scope_ascom.cpp
```diff
@@ -190,11 +196,13 @@ void ScopeASCOM::SetupDialog()
...
         }
     }
-    // destroy the COM object now as this reduces the likelhood of getting into a
+    // If we just created the COM object for doing the SetupDialog (device not already connected),
+    // we need to destroy it now as this reduces the likelhood of getting into a
     // state where the user has killed the ASCOM local server instance and PHD2 is
     // holding a reference to the defunct driver instance in the global interface
     // table
-    m_gitEntry.Unregister();
+    if (!prevRegistered)
```
agalasso, line 204: let's add a comment explaining why this is conditional:
```suggestion
    // Only un-register from the GIT if we happened to have done the registration here
    // in this function call. Otherwise, if the object was already registered at the time
    // we came into this function, de-registration will be taken care of elsewhere.
    if (!prevRegistered)
```

scope_ascom.cpp
```diff
@@ -149,7 +149,6 @@ bool ScopeASCOM::Create(DispatchObj& obj)
         }
 
         Debug.Write(wxString::Format("Create ASCOM Scope: choice '%s' progid %s\n", m_choice, s_progid[m_choice]));
-
```
agalasso, line 152: cosmetic: can we not not remove this blank line -- I find it helpful to separate the Debug.Write noise from the actual code.

scope_ascom.cpp
```diff
@@ -190,11 +190,11 @@ void ScopeASCOM::SetupDialog()
             wxMessageBox(msg, _("Error"), wxOK | wxICON_ERROR);
         }
     }
-    // destroy the COM object now as this reduces the likelhood of getting into a
-    // state where the user has killed the ASCOM local server instance and PHD2 is
-    // holding a reference to the defunct driver instance in the global interface
-    // table
```
agalasso, line 196: it's good that we added the new comment explaining why the de-registration is conditional, but I think we should also keep this comment explaining why we do the unresgtation

agalasso (commented):
looks good, nice and simple now!

agalasso (approved):

PR #1126, opened 2023-11-29
MacOS: update SVB dylib path patching to match the new SVB SDK dylib files

agalasso:


bretmckee (approved):
Looks like a trivial change.

bwdev01 (approved):

PR #1127, opened 2023-12-01
Fix PHD2 compile on Windows

knro:
This fixes the issue on Windows and PHD2 runs and connect to INDI server just fine. The problem is that add_external project does not appear to be working as intended on Linux.


agalasso (approved):
excellent, thanks @knro !

I'll merge this to the https://github.com/OpenPHDGuiding/phd2/tree/knro-indi_work branch and do some testing on Windows, Mac, and Linux, then merge the branch to master and make a new release of PHD2.

PR #1089, opened 2023-08-11
Migrate to INDI v2.0.0+

knro:
Migrate PHD2 to utilize INDI v2

+ Make libnova required on Linux/MacOS since it is required by INDI. Missing libnova would break some functionality like Calibration Assistant due to missing LST.
+ Use INDI v2.0.3 as base for bundled INDI.
+ Need to produce Windows library
+ Needs testing on MacOS

PR #1129, opened 2023-12-04
INDI client 2.0

agalasso:
Updates PHD2's INDI client to INDI 2.0.

Work done by @knro via #1127 and #1089

Tested on Window, Linux, and Mac


bwdev01 (approved):

scope_indi.cpp
```diff
@@ -51,95 +51,103 @@
...
+        {
+            return coord_prop ? true : false;
+        }
+        bool   CanSlewAsync() override;
+        bool   CanCheckSlewing() override
+        {
+            return coord_prop ? true : false;
+        }
+
+        double GetDeclination();
```
jpaana, line 142: This method renaming to GetDeclination() back from GetDeclinationRadians() breaks declination compensation as PHD2 doesn't get the correct declination anymore. Renaming this function back to GetDeclinationRadians() fixes it.

jpaana: Sure, made that in #1135

PR #1145, opened 2023-12-26
fix cmake error on Windows

agalasso:
The commit for #1144 bumped the cmake minimum version to 3.24 since we now use the FetchContent module.  This had a side-effect of changing policy CMP0126 https://cmake.org/cmake/help/latest/policy/CMP0126.html which caused the wxWidgets find_package call to fail as the wxWidgets_CONFIGURATION variable was being set as a normal (non-cache) valriable.


agalasso:
I'm going to merge this without review as without this fix Windows builds are broken.  @bwdev01 if you have any requested changes I can do a followup pr.

bwdev01:
Looks good, thanks.

PR #1156, opened 2023-12-28
Fix crash due to inaccessible pointer

knro:
Simple fix for a crash in profile wizard


bwdev01:
Good catch, thanks for fixing this!

agalasso (approved):
thanks!

PR #1159, opened 2023-12-30
Better error message for RGB FITS file from INDI

### LINKED ISSUE #1157: Using Open Astro Guider (IMX290) with PHD2
I will come up with some proposed patches to the code at some point but for now 
ISSUE / FIX1 : here is a suggestion that I mentioned in https://groups.google.com/g/open-phd-guiding/c/vVk9Qa6H7XI/m/QAJWXJ1XAAAJ - to my mind this is very generic and might apply to any colour web cam

The problem "Unsupported type or read error loading FITS file" reported and no camera connection can be made to PHD2.

looking at phd2/cam_indi.cpp .. the fits file produced by Open Astro Guider and INDI V4L2 CCD driver for the IMX290 chip has NAXIS = 3 / number of data axes .. so the phd2 code can't handle this type of image.   A simple modification to phd2/cam_indi.cpp fixes this:
`-    if (nhdus != 1 || naxis != 2)
+    if (nhdus < 1 || naxis < 1)
     {
         pFrame->Alert(_("Unsupported type or read error loading FITS file"));
         PHD_fits_close_file(fptr);
`
I do not know what other implications this change has but it did allow me to get the camera looping to work in PHD2.

ISSUE / POSSIBLE_FIX2 : Apparently there are two meridian options.. One measuring angles from 0 degrees (I am based in London so this is what I would expect) and the other measuring angles from 180 degrees (I guess is I lived in Hawaii that is what I would expect).  This causes a bit of a problem with the Open Astro Guider... see https://wiki.openastrotech.com/en/Knowledge/Software#useful-python-scripts 

> there is some mismatch between Meade protocol implementation in generic LX200 GPS INDI driver for Linux and OAT ASCOM driver for Windows

> OAT expects 179W is 359 and 179E is 1 from 180 at Greenwich.... LX200 sends 179W is -179 and 179E is 179 from 0 at Greenwich

I am working on this second issue at the moment.. 
My immediate thought was to use Meridian offset (degrees) = -179 ..https://openphdguiding.org/man/Tools.htm#Calibration_Assistant
However this number is restricted to a maximum of 50 degrees.  The result of this is when I press the Slew button on the calibration assistant page the mount slews to the North and attempts to Az 180 DE 0 that is actually pointing into the ground.

The problem "Can't press the SLEW button on  the calibration assistant page without the Open Astro Tracker crashing into its supports"

My POSSIBLE_FIX2 is to modify Meridian offset (degrees) to allow setting 180 degrees to handl this apparent mismatch.

Open to any comments suggestions / other alternative configurations.

pchev:
Change the error message when receiving a FITS file from INDI with NAXIS=3
Please comment if the new message is not clear enough.

Fix #1157

### DIFF
--- a/cam_indi.cpp
+++ b/cam_indi.cpp
@@ -864,9 +864,15 @@ bool CameraINDI::ReadFITS(CapturedFrame *frame, usImage& img, bool takeSubframe,
     int nhdus = 0;
     fits_get_num_hdus(fptr, &nhdus, &status);
 
+    if (naxis == 3)
+    {
+        pFrame->Alert(_("RGB images are not supported, please switch the INDI driver to Mono"));
+        PHD_fits_close_file(fptr);
+        return true;
+    }
     if (nhdus != 1 || naxis != 2)
     {
-        pFrame->Alert(_("Unsupported type or read error loading FITS file"));
+        pFrame->Alert(_("Unsupported FITS file"));
         PHD_fits_close_file(fptr);
         return true;
     }


agalasso (approved):
perfect, thank you!

PR #1162, opened 2024-01-26
Enhance contrast for very low contrast images.

Eyeke2:
This adjustment prevents the frames from being displayed as all-white, which is a visually unpleasant artifact commonly encountered in very dark frames without visible stars. By slightly increasing the maximum value when the difference between min and max values is minimal, we improve the visibility of image details while avoiding the all-white display issue.

### DIFF
--- a/guider.cpp
+++ b/guider.cpp
@@ -441,6 +441,15 @@ bool Guider::PaintHelper(wxAutoBufferedPaintDCBase& dc, wxMemoryDC& memDC)
         {
             int blevel = m_pCurrentImage->FiltMin;
             int wlevel = m_pCurrentImage->FiltMax;
+
+            // Enhance contrast for very low contrast images. This adjustment prevents the frames
+            // from being displayed as all-white, which is a visually unpleasant artifact commonly
+            // encountered in dark frames. By slightly increasing the maximum value
+            // when the difference between min and max values is minimal, we improve the visibility
+            // of image details while avoiding the all-white display issue.
+            if (wlevel < 4096 && (wlevel - blevel) < 16)
+                wlevel += 16;
+
             m_pCurrentImage->CopyToImage(&m_displayedImage, blevel, wlevel, pFrame->Stretch_gamma);
         }
 


agalasso:
This seems promising, thanks for the PR.  Could you suggest some way to demonstrate the effect of the change so we can see the effect?  For example, you could attach a camera image (file > save image), then reviewers could build the branch and test with the camera image to see the improvement.

Eyeke2:
> This seems promising, thanks for the PR. Could you suggest some way to demonstrate the effect of the change so we can see the effect? For example, you could attach a camera image (file > save image), then reviewers could build the branch and test with the camera image to see the improvement.

Unfortunately, my QHYIII290 camera which I used in testing, where this "white screen" happened is now broken. I will try to create a simulated camera image which will show the effect and attach the file (it will require ASCOM camera simulator which is able to load an image file).

Eyeke2:
Actually, I'm not sure if this is sufficient for fixing the white-screen issue for good. It's still possible to generate a dark image with very small difference between min and max pixel values to cause PHD2 to show all-white screen (or nearly all-white), even when screen gamma is set to the rightmost position. If you have other idea how to fix it, I'd like to know. I have addressed it for my specific camera, and haven't thought about more general solution.

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_BOX_AVERAGE - using SkySimulator ASCOM
image

Eyeke2:
For scaling factor > 1 I prefer wxIMAGE_QUALITY_BILINEAR - it shows overall acceptable quality in main window and also in the star profile window. Next I will be posting screenshots for scaling factor < 1.

Eyeke2:
A screenshot with scaling factor of 0.75 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 0.99 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

PR #1165, opened 2024-01-26
Minor speed optimizations

Eyeke2:
Includes speed optimization with insignificant impact on image quality when performing rescaling of captured camera image on the main window.

### DIFF
--- a/guider.cpp
+++ b/guider.cpp
@@ -487,7 +487,7 @@ bool Guider::PaintHelper(wxAutoBufferedPaintDCBase& dc, wxMemoryDC& memDC)
 
                     if (newWidth > 0 && newHeight > 0)
                     {
-                        m_displayedImage->Rescale(newWidth, newHeight, wxIMAGE_QUALITY_HIGH);
+                        m_displayedImage->Rescale(newWidth, newHeight, wxIMAGE_QUALITY_BILINEAR);
                     }
                 }
             }


usImage.cpp
```diff
@@ -221,8 +221,7 @@ static unsigned char *buildGammaLookupTable(int blevel, int wlevel, double power
     if (blevel > 0xffff) blevel = 0xffff;
     if (wlevel > 0xffff) blevel = 0xffff;
 
-    for (int i = 0; i <= blevel; ++i)
-        result[i] = 0;
+    memset(result, 0, blevel + 1);
```
agalasso, line 224: I suggest reverting these memsets. They introduce some fragility in implicitly assuming sizeof(*result) is 1. I could be wrong but this feels more like a personal preference of using memset over an explicit loop. Did you really see any measurable difference from this change?

usImage.cpp
```diff
@@ -231,8 +230,8 @@ static unsigned char *buildGammaLookupTable(int blevel, int wlevel, double power
         result[i] = pow(d, (float) power) * 255.0;
     }
 
-    for (int i = wlevel; i < 0x10000; ++i)
-        result[i] = 255;
+    if (wlevel < 0x10000)
+        memset(result + wlevel, 255, 0x10000 - wlevel);
```
agalasso, line 234: suggest reverting this change

guider.cpp
```diff
@@ -487,7 +487,7 @@ bool Guider::PaintHelper(wxAutoBufferedPaintDCBase& dc, wxMemoryDC& memDC)
 
                     if (newWidth > 0 && newHeight > 0)
                     {
-                        m_displayedImage->Rescale(newWidth, newHeight, wxIMAGE_QUALITY_HIGH);
+                        m_displayedImage->Rescale(newWidth, newHeight, wxIMAGE_QUALITY_BILINEAR);
```
agalasso, line 490: > insignificant impact on image quality

It would be really helpful to see a picture of what the difference looks like in the PR description.  We'd like to take your word for it but a picture would be better!

usImage.cpp
Eyeke2: It's a micro optimization, I agree to revert it.

guider.cpp
Eyeke2: A screenshot before this change
old

Eyeke2: A screenshot after the change - demonstrates sharper details in the starfield.

Eyeke2:
There is a side-effect of this fix, which makes rescaled images sharper. The PR not only saves CPU time but it also helps to visually differentiate between bad pixels and stars, which due to previously used high quality rescaling makes hot pixels look more like tiny stars than what they really are - hot pixels. Here is a screenshot using camera simulator with 1 star and 10 hot pixels - before the fix.
old-sim-with-hotpixels

Eyeke2:
A screenshot with 10 hot pixels and 1 star - showing the effect of PR.
new-sim-with-hotpixels

agalasso:
> rescaling makes hot pixels look more like tiny stars than what they really are - hot pixels

actually, if you click the star profile image it will toggle to raw mode (uninterpolated) which is the best way to see the zoomed-in raw pixels.

But anyway, I do think this would be a good change for the case of scaling down large frames.

Did you try wxIMAGE_QUALITY_NEAREST? That is actually by far the fastest option.  Here are some timings I captured (scaling down a large frame):

box_average = high = 40ms
nearest neighbor = 2ms
bilinear = 26ms

Personally I prefer nearest neighbor when scaling up a small image as it makes the hot pixels more visible like you said.  But scaling down is what this pr is about.  Maybe we should use bilinear when scaling down and nearest neighbor when scaling up? (scaling up a small image should be somewhat rare these days with current generation guide cameras).  But if we could get away with nearest neighbor for resizing down if the quality is adequate, that would be by far the fastest.

Eyeke2:
At first I thought to use nearest neighbor, but for certain types of images it results in stars being displayed pixelized (which can be disturbing to the folks used to see naturally looking smooth star shapes). Box average takes longer than suggested bilinear, but I have overlooked scaling down. I will consider and do some tests in the next days to evaluate this option. Scaling up is not so rare, especially after switching to the 4K screen.

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_BILINEAR - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_BICUBIC - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 1.5 and quality set to wxIMAGE_QUALITY_HIGH - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 0.6 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 0.75 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
A screenshot with scaling factor of 0.99 and quality set to wxIMAGE_QUALITY_NEAREST - using SkySimulator ASCOM
image

Eyeke2:
@agalasso based on the supplied screenshot samples I would suggest to modify this PR to use wxIMAGE_QUALITY_NEAREST (0) for scaling factors < 1 and use wxIMAGE_QUALITY_BILINEAR (1) for scaling factors > 1.

agalasso:
@Eyeke2 thanks for the screenshots. That sounds reasonable. Do you want to update the PR as proposed?

Eyeke2:
@agalasso I still have doubts about using wxIMAGE_QUALITY_NEAREST for scaling factor < 1. After testing it on another computer with a small screen and seeing pixelization it creates when down scaling I tend to leave it always set to wxIMAGE_QUALITY_BILINEAR. Here is some background information which could be helpful for making the decision:

In wxWidgets, when downscaling an image and considering a balance between quality and CPU usage, wxIMAGE_QUALITY_BILINEAR is indeed a recommended option. Bilinear filtering provides a good compromise between the higher computational cost of more sophisticated algorithms (like bicubic resampling) and the lower quality but faster performance of nearest neighbor (wxIMAGE_QUALITY_NEAREST) resampling.

Understanding Rescaling Options:
wxIMAGE_QUALITY_NEAREST: This method is the fastest as it simply picks the nearest pixel color to the target position without any interpolation. The result is often pixelated or blocky, especially noticeable in downscaling.

wxIMAGE_QUALITY_BILINEAR: Bilinear filtering averages the colors of the nearest 2x2 neighborhood of pixel squares, providing smoother transitions between pixels compared to nearest neighbor. It strikes a good balance between image quality and processing time, making it suitable for many applications where moderate downscaling quality is acceptable without significant CPU load.

wxIMAGE_QUALITY_BICUBIC (if available) and wxIMAGE_QUALITY_HIGH (typically maps to bicubic or better): These methods offer higher quality by considering more pixels in the interpolation process, which results in smoother images but at a higher computational cost.

bwdev01:
I'm not sure where we're going with this.  IMO, none of this is important in terms of actual guiding, it's cosmetic.  But what *is* important is that we don't slow down the image_ready processing.  We have people with strain-wave mounts who need to do fast-cadence guiding and we don't want to generate a bunch of new complaints about the processing overhead in PHD2.  Let's not fix something that isn't broken in a way that matters, and if we do make changes, let's be sure we're looking at real star images from a real guiding system. :-)

agalasso:
> I'm not sure where we're going with this. IMO, none of this is important in terms of actual guiding, it's cosmetic. But what _is_ important is that we don't slow down the image_ready processing.

@bwdev01 I agree 100%.   I posted some timings in one of the earlier comments ([here](https://github.com/OpenPHDGuiding/phd2/pull/1165#issuecomment-1913327214)). The current value, wxIMAGE_QUALITY_HIGH (alias for wxIMAGE_QUALITY_BOX_AVERAGE) is slower than the proposed change, so if we can get away with with nearest neighbor and/or bilinear it will the reduce cpu cost of image_ready processing in the case of scaling up or scaling down the frame for display.

bwdev01:
Ok, thanks for the clarification.

Eyeke2:
@bwdev01 @agalasso I admit the name of this PR is now misleading, it should say something like "improve speed of image rescaling while making hot pixels less resembling stars". For big 4k screens and  cameras with large pixel count this fix is quite noticeable in terms of UI responsiveness (which is even more pronounced in debug mode). The highest rescaling quality used before this PR is an overkill which easily causes hot pixels to be rendered with profiles reminiscent of small stars.

Eyeke2:
@agalasso I removed the memset optimizations and left only the rescale quality fix.

agalasso (approved):
@bwdev01 could you review?

bwdev01 (approved):

PR #1167, opened 2024-01-28
Star Profile Window: limit size of the large font.

Eyeke2:
As a follow-up commit to Star Profile tweaks, this is designed to put an upper limit on large font size and prevent making it too big.


Eyeke2:
A side-by-side screenshots comparing before and after the fix.
before
after

Eyeke2:
For additional clarity, here is a screenshot using version before recent merge of Star Profile tweaks to master.
old

agalasso (approved):

PR #1168, opened 2024-02-04
Prevent resizing of rows and columns in stats window

Eyeke2:
Prevent resizing of rows and columns in stats grids to protect against inadvertent change of cell size by the user.

### DIFF
--- a/statswindow.cpp
+++ b/statswindow.cpp
@@ -82,6 +82,7 @@ StatsWindow::StatsWindow(wxWindow *parent)
     m_grid1->SetCellValue(1, 1, "");
     m_grid1->SetCellValue(1, 2, "");
     m_grid1->ClearSelection();
+    m_grid1->DisableDragGridSize();
 
     m_grid2 = new wxGrid(this, wxID_ANY);
     m_grid2->CreateGrid(12, 2);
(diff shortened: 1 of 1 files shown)


agalasso (commented):
It does make sense to disable resizing the cells.  We should be able to achieve the same effect more simply by calling [DisableDragGridSize](https://docs.wxwidgets.org/3.0/classwx_grid.html#a8becb571ff4b0b23e69f0397eeaa391a) on the two grids.

Eyeke2:
@agalasso I have overlooked DisableDragGridSize(). I've tested it and it indeed works as intended. Simpler solutions are surely better. Should I force push your suggestion over this branch for PR?

agalasso:
> Should I force push your suggestion over this branch for PR?

It's up to you whether you want to to force-push or push as a separate commit.   Either way it will make no difference i the end as we will always squash PRs merged to master and the history on master will always show a single commit for the pr.

Eyeke2:
> > Should I force push your suggestion over this branch for PR?
> 
> It's up to you whether you want to to force-push or push as a separate commit. Either way it will make no difference in the end as we will always squash PRs merged to master and the history on master will always show a single commit for the pr.

@agalasso - updated this PR to use DisableDragGridSize() as you suggested. Thanks!

agalasso (approved):

PR #1172, opened 2024-02-11
Set  LIBNOVA  when using bundled Indi

nrackwit:
In thirdparty.cmake also set the LIBNOVA compiler option when compiling indi_scope.cpp for PHD2 against bundled the Indi client. The compiler flag is already being set when compiling for PHD2 against system the system's Indi client.

This helps ensure that the getCoordinates implementation takes sidereal time into account for providing a correct RA coordinate.


pchev:
For me this look good.

Can you confirm this fix your issue : https://groups.google.com/g/open-phd-guiding/c/gpfSCkkbgH0/m/ZNMRSpJuAwAJ

nrackwit:
Yes, this fixes [Some odd RA coordinate behavior in the 2.6.13dev1 development build (with INDI)](https://groups.google.com/g/open-phd-guiding/c/gpfSCkkbgH0/m/ZNMRSpJuAwAJ).

I tested with either option -DUSE_SYSTEM_LIBINDI=OFF and -DUSE_SYSTEM_LIBINDI=ON and both builds now generate the same RA coordinate value. 

The Drift Align tool now also shows the correct 'Offset to Meridian' value.

bwdev01:
Nice work!  Thanks for figuring this out.

 

Bruce

agalasso (commented):
@nrackwit thanks for fixing this!  I think the LIBNOVA definition can go outside the `endif()` on line 637 so that Macs get the fix too, in addition to Linux.
I just tested building on Mac with the definition outside the endif and it works.
Do you want to make that change on this pr?

nrackwit:
@agalasso - yes, sure thing, just give me a bit to get the pr updated. Thank you for confirming the Mac part to this change!

nrackwit:
Alright, the PR is amended, it should be all set now.

agalasso (approved):
thanks!

Issue #1174, opened 2024-02-13
Problem with connection status when connecting Indi devices in v2.6.13dev1


nrackwit:
When I connect an IndiCamera and IndiMount the PHD2.6.13dev1 the Equipment Connection Dialog elements and status  information seem out of sync with true connection status. I originally noticed this when connecting my real equipment but it works just as well using Indi's camera and mount simulators. 
 
Upon connecting camera and mount via the "Connect All" button PHD2 generates an alert that "You must connect a camera before loading a dark library" although the profile has a corresponding dark library and the camera seems to be connected successfully. I get the "Disconnect" button and the "All devices Button" is green.  The "Darks" status shows red, though. The "Connect" button for the mount is still available and when clicking it the mount connection seems to get established. However the status bar shows "Mount connection failed".

When disconnecting ("Disconnect All") and reconnecting ("Connect All") the equipment the dark library now is found but now there also is an alert that "You must connect a camera before loading a bad-pixel map" (not that I have one, but I think that PHD2 usually doesn't require one - so the problem seems to be that PHD2 thinks there is no camera connection)

From the PHD2 log it seems that PHD2 tries to initialize the camera properties before the Indi connection is actually ready, I.e: 

> 19:28:18.363 00.001 548548464704 Connecting to camera [INDI Camera [CCD Simulator]] id = []
> 19:28:18.363 00.000 548548464704 INDI Camera connecting to device [CCD Simulator]
> 19:28:18.368 00.005 548548464704 GetDouble("/profile/1/camera/pixelsize", 0.000000) returns 5.200000
> 19:28:18.369 00.001 548548464704 DoConnectCamera: reconnecting=0 warningIssued=0 lastCam=[INDI Camera [CCD Simulator]] scaleRatio=1.000
> **19:28:18.369 00.000 548548464704 Connected Camera: INDI Camera [CCD Simulator]**
> 19:28:18.369 00.000 548548464704 FullSize=(0,0)
> 19:28:18.369 00.000 548548464704 PixelSize=5.20
> 19:28:18.369 00.000 548548464704 BitsPerPixel=0
> 19:28:18.369 00.000 548548464704 HasGainControl=0
> 19:28:18.369 00.000 548548464704 HasShutter=0
> 19:28:18.370 00.001 548548464704 HasSubFrames=1
> 19:28:18.370 00.000 548548464704 ST4HasGuideOutput=0
[...]

PR #1177, opened 2024-02-14
Update INDI device's Connect method to wait for 'Connected' property

nrackwit:
This push attempts to fix "Problem with connection status when connecting Indi devices in v2.6.13dev1 [#1174 ]"

This new code monitors the INDI device's 'Connected' property and waits for it to be 'true' before releasing the Connect method back to the UI. I tried to move the wait loop into the RunInBG template so users are able to cancel.

It's been tested with an aarch64 bit build using the Indi scope and camera simulators. I verified that
* Camera simulator and mount simulator are connecting
* Both devices show as connected
* The Dark frame library is assigned successfully
* No alerts with respect to camera connectivity and bad-pixel maps or dark libraries 
* "Connect All" button connects both devices and then closes gear selection dialog.

Concerns or advise is welcome. I am not overly familiar with the overall device management code base so there is a good chance that this change is overlooking something or not following best practices - let me know!

A debug log for the test is attached:
[PHD2.log](https://github.com/OpenPHDGuiding/phd2/files/14271621/PHD2.log)

### DIFF
--- a/scope_indi.cpp
+++ b/scope_indi.cpp
@@ -298,7 +298,49 @@ bool ScopeINDI::Connect()
     // Receive messages only for our mount.
     watchDevice(INDIMountName.mb_str(wxConvUTF8));
 
-    return !connectServer();
+    Debug.Write(wxString::Format("Waiting for 30s for [%s] to connect...\n", INDIMountName));
+
+    ///* Wait in foreground for driver to establish a device connection */
+    //if (connectServer())
+    //{
+    //    Debug.Write(wxString::Format("Waiting for 30s for [%s] to connect...\n", INDIMountName));
+    //    int i = 0;
+    //    while (!Connected && i++ < 300) 
+    //        wxMilliSleep(100);
+    //}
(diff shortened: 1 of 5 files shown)


scope_indi.cpp
```diff
@@ -298,7 +298,49 @@ bool ScopeINDI::Connect()
...
+    //{
+    //    Debug.Write(wxString::Format("Waiting for 30s for [%s] to connect...\n", INDIMountName));
+    //    int i = 0;
+    //    while (!Connected && i++ < 300) 
+    //        wxMilliSleep(100);
+    //}
+    //
+    //// We need to return FALSE if we are successful???
+    //return !Connected;
+
```
agalasso, line 314: let's remove the commented-out lines
```suggestion
```

agalasso (commented):
looks good!

nrackwit:
Thank you for reviewing, Andy!

agalasso (approved):

PR #1179, opened 2024-03-09
Mac: workaround for wxWidgets crash displaying wxInfoBar icon

agalasso:


bwdev01 (approved):
Isn't it the case that the only possible icon is a giant exclamation mark?  That may be entirely useless, should we just nuke that on all the platforms?

agalasso:
⚠️ Good point ;)

I see a couple callers that use wxICON_INFORMATION rather than the default, wxICON_EXCLAMATION, so we should take a look at how those look as well.  I would not be opposed to removing them unconditionally in another pr.

Issue #1180, opened 2024-03-14
Add event server interfaces for 'variable delay' feature


bwdev01:
User has requested event server interfaces to fine-tune variable delay properties for use with absolute encoder mounts.

PR #1183, opened 2024-03-25
Add event server interfaces to get/set variable delay parameters. Fix…

bwdev01:
…es #1180.  Param interface uses units of seconds to match AD UI.

### DIFF
--- a/event_server.cpp
+++ b/event_server.cpp
@@ -1861,6 +1861,36 @@ static void get_settling(JObj& response, const json_value *params)
     response << jrpc_result(settling);
 }
 
+static void get_variable_delay_settings(JObj& response, const json_value* params)
+{
+    JObj rslt;
+
+    VarDelayCfg delayParams = pFrame->GetVariableDelayConfig();
+    rslt << NV("Enabled", delayParams.enabled)
+        << NV("ShortDelay", delayParams.shortDelay / 1000)
+        << NV("LongDelay", delayParams.longDelay / 1000);
+    response << jrpc_result(rslt);
+}
+
+static void set_variable_delay_settings(JObj& response, const json_value* params)
+{
+    Params p("Enabled", "ShortDelay", "LongDelay", params);
+    const json_value* p0 = p.param("Enabled");
+    const json_value* p1 = p.param("ShortDelay");
+    const json_value* p2 = p.param("LongDelay");
+    bool enabled;
+    double shortDelay;
+    double longDelay;
+    if (!p0 || !p1 || !p2 || !bool_param(p0, &enabled) || !float_param(p1, &shortDelay) || !float_param(p2, &longDelay))
+    {
+        response << jrpc_error(JSONRPC_INVALID_PARAMS, "expected Enabled, ShortDelay, LongDelay params)");
+        return;
+    }
+    VarDelayCfg currParams;
+    pFrame->SetVariableDelayConfig(enabled, (int)shortDelay * 1000, (int)longDelay * 1000);
+    response << jrpc_result(0);
+}
+
 static GUIDE_DIRECTION dir_param(const json_value *p)
 {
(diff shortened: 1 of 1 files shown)


event_server.cpp
```diff
@@ -1861,6 +1861,36 @@ static void get_settling(JObj& response, const json_value *params)
...
+    bool enabled;
+    double shortDelay;
+    double longDelay;
+    if (!p0 || !p1 || !p2 || !bool_param(p0, &enabled) || !float_param(p1, &shortDelay) || !float_param(p2, &longDelay))
+    {
+        response << jrpc_error(JSONRPC_INVALID_PARAMS, "expected Enabled, ShortDelay, LongDelay params)");
+        return;
+    }
+    VarDelayCfg currParams;
+    pFrame->SetVariableDelayConfig(enabled, (int)shortDelay * 1000, (int)longDelay * 1000);
```
agalasso, line 1890: better do the multiplication before the cast to int!
```suggestion
    pFrame->SetVariableDelayConfig(enabled, (int)(shortDelay * 1000.), (int)(longDelay * 1000.));
```

bwdev01: Thanks.  I've posted the update, looking for another review.

event_server.cpp
```diff
@@ -1861,6 +1861,36 @@ static void get_settling(JObj& response, const json_value *params)
...
 }
 
+static void get_variable_delay_settings(JObj& response, const json_value* params)
+{
+    JObj rslt;
+
+    VarDelayCfg delayParams = pFrame->GetVariableDelayConfig();
+    rslt << NV("Enabled", delayParams.enabled)
+        << NV("ShortDelay", delayParams.shortDelay / 1000)
+        << NV("LongDelay", delayParams.longDelay / 1000);
```
agalasso, line 1871: I'm guessing the intent is to output fractional seconds (floating point), not just integer seconds, right?
if so, we should make these into floating point calculations:
```suggestion
        << NV("ShortDelay", (double) delayParams.shortDelay / 1000., 3)
        << NV("LongDelay", (double) delayParams.longDelay / 1000., 3);
```

agalasso (commented):
oops, missed this issue on the first review

bwdev01:
Well, the AD deals with integer values in the UI and given the context, there’s no real need for decimal values.  However, the parsing machinery in place in EventServer doesn’t distinguish between integers and floating point.  So we either have to enforce integer input params or accept and return floating point – which is what your change will do.   Is that how you’d like to handle it?  Are there other API entry points that enforce integer input?

event_server.cpp
```diff
@@ -1861,6 +1861,36 @@ static void get_settling(JObj& response, const json_value *params)
...
+    bool enabled;
+    double shortDelay;
+    double longDelay;
+    if (!p0 || !p1 || !p2 || !bool_param(p0, &enabled) || !float_param(p1, &shortDelay) || !float_param(p2, &longDelay))
+    {
+        response << jrpc_error(JSONRPC_INVALID_PARAMS, "expected Enabled, ShortDelay, LongDelay params)");
+        return;
+    }
+    VarDelayCfg currParams;
+    pFrame->SetVariableDelayConfig(enabled, (int)(shortDelay * 1000.), (int)(longDelay * 1000.));
```
agalasso, line 1890: now that I understand these delays are always positive integer number of seconds, I think we should change this back to the way you had it before since this way would now allow values that are millisecond resolution and we do not want to allow that.
I would suggest adding "Seconds" to the API parameter names and variable names so that it is clearer that we're dealing with seconds (and not milliseconds like all other durations in the code base)
```suggestion
    pFrame->SetVariableDelayConfig(enabled, (int) shortDelaySeconds * 1000), (int) longDelaySeconds * 1000));
```

agalasso (approved):

PR #1185, opened 2024-04-04
UpdateCurrentDistance: use distanceRA for calculating RA averages

Eyeke2:
Fix incorrect calculation of average distances for RA


agalasso (approved):
thanks @Eyeke2 !

PR #1186, opened 2024-04-09
Histogram builder: fix incorrect delete operator

Eyeke2:
Using delete instead of delete[] for memory allocated with new[] is a common mistake that can lead to undefined behavior. The consequences of this mismatch between allocation and deallocation mechanisms can vary depending on the compiler, the runtime environment, and the specifics of the program.


agalasso (approved):

agalasso:
@Eyeke2 thanks for the pr

PR #1188, opened 2024-04-15
opencv - update to version 4

agalasso:
Testing:
 - built installer, installed and tested opencv camera on Windows


bwdev01 (approved):

pchev:
I tested compilation on Linux without issue.

Later today I will try if cam_opencv can work on Linux.

PR #1191, opened 2024-04-16
Add OpenCV camera with Linux, the new code for OpenCV4 work fine

pchev:
With the change for OpenCV4 the OpenCV camera now work fine on Linux.
Tested with a laptop webcam.

This is interesting to add because this can support other devices than the V4L2 INDI driver. 

The change is trivial, but please let me know if there is another/better location for "find_package".
I not added conditional switch because OpenCV will anyway be required by the new planetary module.


agalasso (approved):
Thanks Patrick!

PR #1192, opened 2024-04-18
minor refactor for code reuse

agalasso:
move some binning methods out of cam_touptek.cpp into image_math.h so the methods can be used by another camera in a forthcoming PR

no functional changes

### DIFF
--- a/cam_touptek.cpp
+++ b/cam_touptek.cpp
@@ -3,7 +3,7 @@
  *  Open PHD Guiding
  *
  *  Created by Andy Galasso
- *  Copyright (c) 2018 Andy Galasso.
+ *  Copyright (c) 2018-2024 openphdguiding.org
  *  All rights reserved.
  *
  *  This source code is distributed under the following "BSD" license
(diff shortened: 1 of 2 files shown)


bwdev01 (approved):
Since this does software binning, should the capability be exposed for cameras that don't do hardware binning?  This would affect the new-profile-wizard and the AD.

agalasso:
> should the capability be exposed for cameras that don't do hardware binning

good point. this _could_ be used by cameras that do not do hardware binning -- we can do that as separate prs.  Let me know if you had any specific cameras in mind and we can add it.

bwdev01:
It was just a general notion, not specific to a particular camera.  I thought we had an outstanding feature request from way back for this but maybe I’m mistaken.  Having not spent much time thinking about it, perhaps we would offer the software binning option in the UI for any camera that reports a max-binning value of zero. With big CMOS sensors on guide cameras, this could be a useful way to reduce image sizes and download times.  But not anything that should affect this PR.

PR #1200, opened 2024-04-21
Target window: tweak positions of graphical elements to avoid overlap (part4 for planetary guiding)

Eyeke2:
I'll add later a few screenshots to show what was fixed.

### DIFF
--- a/target.cpp
+++ b/target.cpp
@@ -337,7 +337,7 @@ void TargetClient::OnPaint(wxPaintEvent& WXUNUSED(evt))
 
     // Draw labels
     int d_width, d_height;
-    dc.GetTextExtent(_("0"), &d_width, &d_height);
+    dc.GetTextExtent(_T("0"), &d_width, &d_height);
     dc.DrawText(_("RA"), leftEdge, center.y - 3 - d_height);
     dc.DrawText(_("Dec"), center.x + 5, center.y - radius_max - d_height);
 
--- a/target.cpp
+++ b/target.cpp
@@ -336,8 +336,10 @@ void TargetClient::OnPaint(wxPaintEvent& WXUNUSED(evt))
     }
 
     // Draw labels
-    dc.DrawText(_("RA"), leftEdge, center.y - 15);
-    dc.DrawText(_("Dec"), center.x - 35, topEdge - 3);
+    int d_width = 10, d_height = 10;
+    dc.GetTextExtent(_("0"), &d_width, &d_height);
+    dc.DrawText(_("RA"), leftEdge, center.y - 3 - d_height);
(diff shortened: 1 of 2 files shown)


Eyeke2:
Before the fix 
image

Eyeke2:
After the fix 
image

target.cpp
```diff
@@ -336,8 +336,10 @@ void TargetClient::OnPaint(wxPaintEvent& WXUNUSED(evt))
     }
 
     // Draw labels
-    dc.DrawText(_("RA"), leftEdge, center.y - 15);
-    dc.DrawText(_("Dec"), center.x - 35, topEdge - 3);
+    int d_width = 10, d_height = 10;
```
agalasso, line 339: minor, these initializers are not needed, let's remove them to avoid confusion
```suggestion
    int d_width, d_height;
```

target.cpp
```diff
@@ -336,8 +336,10 @@ void TargetClient::OnPaint(wxPaintEvent& WXUNUSED(evt))
     }
 
     // Draw labels
-    dc.DrawText(_("RA"), leftEdge, center.y - 15);
-    dc.DrawText(_("Dec"), center.x - 35, topEdge - 3);
+    int d_width = 10, d_height = 10;
+    dc.GetTextExtent(_("0"), &d_width, &d_height);
```
agalasso, line 340: we do not want this string translated
```suggestion
    dc.GetTextExtent(_T("0"), &d_width, &d_height);
```

target.cpp
agalasso: curious, what does the leading `d_` prefix stand for in these variable names?

agalasso (commented):
thanks for the pr, nice little cleanup

Eyeke2:
@agalasso I think that having explicit RA and DEC labels is useful, since it adds information in cases when error distribution becomes asymmetrical and shows more elongation or drift in one axis than another. There are other indicators which are related to pier side orientation, which could be somewhat confusing (like SkyE/SkyW and SkyN/SkyS). This PR just does some tweaking to positions of all text labels to avoid them overlapping with each other as well as with the scale markers.

Eyeke2:
@agalasso, @bwdev01 Adding Bruce to this discussion to get his opinion on this PR, which is one in the series leading up to the planetary guiding feature.

target.cpp
Eyeke2: Agreed. It took me a minute to recall what I meant by d_ - probably a shortcut for 'digit'.

target.cpp
Eyeke2: Ok

Eyeke2:
@agalasso I've approved your updates, please ignore my previous comments.

PR #1203, opened 2024-04-27
Rework of image-scale change handling

bwdev01:
This revises earlier code that reacted to image scale changes in an existing profile.  This revision takes a more straightforward approach that mimics the code in the new-profile-wiz to create from-scratch settings for calibration step-size and MinMoves.  It forces a fresh calibration after the changes are made but doesn't do the broad-brush guide algorithm resets that were done in the original code.  The original code was too vulnerable to situations where user-controlled settings were already out-of-whack.

### DIFF
--- a/advanced_dialog.cpp
+++ b/advanced_dialog.cpp
@@ -551,12 +551,16 @@ void AdvancedDialog::SetPixelSize(double val)
         m_pCameraCtrlSet->SetPixelSize(val);
 }
 
+inline static double SiderealRateFromGuideSpeed(double guideSpeed)
+{
+    double const siderealSecsPerSec = 0.9973;
+    return guideSpeed * 3600.0 / (15.0 * siderealSecsPerSec);
+}
+
 // Get the best estimate for the current mount guide speeds
 double AdvancedDialog::DetermineGuideSpeed()
 {
-    double const siderealSecsPerSec = 0.9973;
     double sidRate = 0.5;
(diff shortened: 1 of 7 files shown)


advanced_dialog.cpp
```diff
@@ -544,47 +551,77 @@ void AdvancedDialog::SetPixelSize(double val)
...
+double AdvancedDialog::DetermineGuideSpeed()
 {
-    LoadValues();           // Insure that control values reflect actual device properties and any changes made outside the AD
-    m_pMountPane->ResetRAGuidingParams();
-    m_pMountPane->ResetDecGuidingParams();
-    UnloadValues();
-}
+    double const siderealSecsPerSec = 0.9973;
+    double sidRate = 0.5;
+    #define RateX(spd)  (spd * 3600.0 / (15.0 * siderealSecsPerSec))
```
agalasso, line 559: minor: better define this as a real function and not a macro. reason: type safety and macro operator precedence problems
```
inline static double SiderealRateFromGuideSpeed(double guideSpeed)
{
    double const siderealSecsPerSec = 0.9973;
    return guideSpeed * 3600.0 / (15.0 * siderealSecsPerSec);
}
```

agalasso (commented):
great work!

advanced_dialog.cpp
bwdev01: Makes sense, thanks.

agalasso (approved):

PR #1204, opened 2024-04-30
Do not try to fetch the Ogma camera library whith OPENSOURCE_ONLY

pchev:
With the recent addition of OGMA camera it is no more possible to build the Ubuntu PPA package.

This is because it is not possible to reach Github from the Ubuntu build server to download the library.
Even with -DOPENSOURCE_ONLY=1 it still try to fetch the library, making the build to failed.

This change just prevent the fetch when OPENSOURCE_ONLY=1.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: phd2
 Section: education
 Priority: optional
 Maintainer: Patrick Chevalley <pch@ap-i.net>
-Build-Depends: debhelper (>= 9), cmake, libwxgtk3.2-dev | libwxgtk3.0-dev | libwxgtk3.0-gtk3-dev, libcfitsio-dev, libopencv-dev, libusb-1.0-0-dev, libudev-dev, libv4l-dev, libnova-dev, libcurl4-gnutls-dev, libindi-dev (>= 2.0), libeigen3-dev, libgtest-dev libopencv-dev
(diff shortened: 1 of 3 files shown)


debian/control
```diff
@@ -2,7 +2,7 @@ Source: phd2
 Section: education
 Priority: optional
 Maintainer: Patrick Chevalley <pch@ap-i.net>
-Build-Depends: debhelper (>= 9), cmake, libwxgtk3.2-dev | libwxgtk3.0-dev | libwxgtk3.0-gtk3-dev, libcfitsio-dev, libopencv-dev, libusb-1.0-0-dev, libudev-dev, libv4l-dev, libnova-dev, libcurl4-gnutls-dev, libindi-dev (>= 2.0), libeigen3-dev, libgtest-dev libopencv-dev
+Build-Depends: debhelper (>= 9), cmake, libwxgtk3.2-dev | libwxgtk3.0-dev | libwxgtk3.0-gtk3-dev, libcfitsio-dev, libopencv-dev, libusb-1.0-0-dev, libudev-dev, libv4l-dev, libnova-dev, libcurl4-gnutls-dev, libindi-dev (>= 2.0), libeigen3-dev, libgtest-dev
```
d33psky, line 5: I assume you did not mean to remove libopencv-dev

d33psky (changes requested):

debian/control
pchev: Yes  this is intended because when I add it I not remark it was already there for a long time, just after libcfitsio-dev.

agalasso (approved):
Thanks Patrick! Sorry I overlooked this when adding OGMA.

debian/control
d33psky: oh intended after all. ok

PR #1210, opened 2024-06-10
Added armv8 (64bit) qhyccd sdk library

ceterumnet:
Adds QHYCCD SDK library for armv8 for Linux


agalasso:
@ceterumnet was the alpaca commit to this pr unintentional?

AstroAir:
Although this PR is a bit confusing, I think it's possible to use nlohmann/JSON to rewrite the current event_server, which can significantly improve code readability?

ceterumnet:
Yes - that was unintentional. I am super confused how that was added to this PR...

ceterumnet:
@agalasso - ok - should be good now.

ceterumnet:
@agalasso - let me know if you need anything else from me on this.

agalasso:
@ceterumnet thanks for the PR.  What SDK version did the armv8 library come from? I'd like to validate that the armv8 library is the same version as the other architectures and that it is the exact same binary distributed by QHY.
Also, we'll need to update the script (build/unpack_qhy_sdk.sh) that we use to unpack the QHY SDK libraries so that the armv8 library gets updated along with the others whenever we update the QHY SDK -- I can make the change to the unpack script as a separate pr once I get the info.

agalasso:
@ceterumnet I put up #1215 which updates PHD2 to use the latest QHY SDK, but also happens to update the build scripts and QHY SDK unpack script to include the arm64 SDK lib.
Could you please check to see if #1215's branch addresses your issue and allows PHD2 to work with your QHY camera on your armv8 system?

ceterumnet:
@agalasso - I think it should work. I have tested it, but there may need to be a small tweak in order for it to build. I'll issue another PR once I've figured that out.

ceterumnet:
@agalasso - FYI - I had no issues building and my camera works now on the Raspberry Pi 5. Just in case anyone ever stumbles across this and used my code...make sure you purge your CMake build directory to avoid any stale values from what I had done.

PR #1215, opened 2024-06-23
QHY cameras: update to SDK version 24.03.29

agalasso:
Update to the current latest SDK from QHY, 24.03.29.

Also updates the build scripts to use the newer QHY SDK packaging layout.

For Linux builds, PHD2 will incorporate QHY's x86_64, arm32, or arm64 SDK library matching the host build architecture; the x86_32, armv6 and armv7 files from the older QHY packaging scheme are no longer used.

Tested with a QHY camera on Windows and Mac.  Confirmed successful build on Linux x86_64.


bwdev01 (approved):
Thanks for doing this and for plowing through the testing.

PR #1227, opened 2024-07-28
Bugfix: prevent reusing a single JsonParser object among all PHD2 clients and across reentrant PHD2 calls.

Eyeke2:
This fix addresses issues arising from sharing a single JsonParser instance, ensuring each client and reentrant call operates independently to avoid conflicts and potential data corruption.

### DIFF
--- a/src/event_server.cpp
+++ b/src/event_server.cpp
@@ -2639,7 +2639,8 @@ void EventServer::OnEventServerClientEvent(wxSocketEvent& event)
     }
     else if (event.GetSocketEvent() == wxSOCKET_INPUT)
     {
-        handle_cli_input(cli, m_parser);
+        JsonParser parser;
+        handle_cli_input(cli, parser);
     }
     else
     {
--- a/src/event_server.h
+++ b/src/event_server.h
@@ -44,7 +44,6 @@ public:
     typedef std::set<wxSocketClient *> CliSockSet;
 
 private:
-    JsonParser m_parser;
     wxSocketServer *m_serverSocket;
     CliSockSet m_eventServerClients;
     wxTimer *m_configEventDebouncer;


agalasso:
> This fix addresses issues

@Eyeke2 is this speculation or did you find any actual issues? If actual issues could you please elaborate on what those issues were?

`OnEventServerClientEvent` is called from the main (UI) thread _only_.  If you found an actual reentrancy scenario, then we probably would have other problems besides just this.

Eyeke2:
@agalasso Yes, the issue happened during the 'set_connected' method. Without the fix, new PHD2 client requests were processed before the completion of 'set_connected'. This resulted in a loss of the original request context at the end of set_connected processing, causing the JSON response to reply with an outdated "id" field in the JSON message.

Here are the specific details of the issue:

The set_connected method can cause the event loop to run reentrantly.
During the execution of 'set_connected', new client requests could be handled before set_connected completed, leading to potential data corruption or context loss.

At the end of the 'set_connected' method, the JSON response sometimes included an old "id" field because the original request context was lost. This behavior indicates a reentrancy problem, where the parser's state was inadvertently shared across different client requests.

Code Warning:

The PHD2 code includes a specific warning about reentrancy in the 'handle_cli_input' function:

static void handle_cli_input(wxSocketClient *cli, JsonParser& parser)
{
    // Bump refcnt to protect against reentrancy.
    //
    // Some functions like set_connected can cause the event loop to run reentrantly. If the
    // client disconnects before the response is sent and a socket disconnect event is
    // dispatched the client data could be destroyed before we respond.
...

agalasso:
@Eyeke2 thank you for providing the details. I put up a PR with a slightly different fix that moves the instantiation of the parser closer to where it is actually used and simplifies the code.
We can close this PR and take #1228 in its place.
Thanks for reporting the issue.

Eyeke2:
@agalasso Sounds great! On a side note, I compiled the most recent master branch and noticed that locals_update.sh seems to be broken. How do you update local language templates?

agalasso:
> @agalasso Sounds great! On a side note, I compiled the most recent master branch and noticed that locals_update.sh seems to be broken. How do you update local language templates?

let's not hijack this thread. you can create an issue https://github.com/OpenPHDGuiding/phd2/issues and we can discuss it there

PR #1228, opened 2024-07-28
event server: fix possible problems when client sends multiple requests

agalasso:
event server: fix possible problems when client sends multiple requests

The `connect` command can cause the event loop to run and for the event server to process additional buffered client commands.  As the JSON parser instance is reused for all client commands, the parser's state will reflect the most recent client command and may lose the state of the unfinished connect command.

The fix is to not reuse the JsonParser instance -- instantiate a dedicated parser for each client command.

Testing:
 - manually exercised the event server by issuing commands with an event server client
 - checked the runtime overhead of instantiating the JsonParser instance per request (negligible, < 100 ns)

### DIFF
--- a/src/event_server.cpp
+++ b/src/event_server.cpp
@@ -2442,8 +2442,13 @@ static bool handle_request(JRpcCall& call)
     }
 }
 
-static void handle_cli_input_complete(wxSocketClient *cli, char *input, JsonParser& parser)
+static void handle_cli_input_complete(wxSocketClient *cli, char *input)
 {
+    // a dedicated JsonParser instance is used for each line of input since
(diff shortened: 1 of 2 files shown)


bwdev01 (approved):
Having a hard time seeing the use-case.  Why would an app spray a stream of messages at the event server?  Isn't this supposed to be an RPC model?

agalasso:
> Why would an app spray a stream of messages at the event server?

Agreed, that's probably why we do not hear about this problem.   But in theory, an app should be allowed to send a multiple (overlapping) requests and then correlate the responses using the request ids -- we made this work with #780 and #781.

Another way to look at this pr is that it is completing the change started by #780.

PR #1233, opened 2024-08-13
Corrected path for installed files

lboclboc:
Some icons and desktop files was references from the src dir but they are located in other places.


agalasso (approved):
thanks!

Issue #1245, opened 2024-10-10
QHY native driver improvements


daleghent:
After some exploratory work, I'm re-titling this issue to be more relevant.

The QHY native driver can use some enhancement to address 1 problem and bring it in line with modern QHY SDK feature sets.

1. Explicitly call `SetQHYCCDBitsMode()` during `Connect()` to set the camera output to either 8 or 16 bit. This would help correct an issue with the saturation level camera configuration item where, in some cases, a saturation of 255 would be enforced despite the camera outputting 16bit data.
2. Add support for QHY cameras with coolers.
3. Add support for specifying an offset. Currently, only the gain is controlled and there is no way to specify an offset. This can result in many pixels becoming clipped, and more will be clipped as gain value increases.
4. Add support for "amplifier noise reduction" feature present on some QHY camera models. This is done by manipulating the `CONTROL_AMPV` camera control.
5. Add support for QHY5II-M row noise reduction option
6. Add support for QHY5III178 / QHY178 high gain mode
7. Optimize exposure+data readout timing based on feedback from Dr. Q @ QHY

daleghent:
WIP branch here https://github.com/daleghent/phd2/tree/qhy-improvements

daleghent:
WIP camera options window

![image](https://github.com/user-attachments/assets/389934d9-1bab-446e-9a64-5698f6912661)

bwdev01:
Hi Dale, this looks good at this point.  My question is whether we understand the peculiar behavior reported regarding the bits-per-pixel property.  In particular, I want to know why we’re constraining the MaxADU value to be 255.  That property is represented by a text field in the UI so the constraining mechanism should be coming from this code when the form is closed:

 

saturationVal = wxMin(val, SaturationValFromBPP(m_pCamera));

 

This suggests that a read of the bits-per-pixel value is returning 8 even though the camera is returning 16-bit data.  I didn’t see anything in your code that did anything in this area although perhaps I missed it.  AFAICT, the m_bpp variable in cam_qhy.cpp is being set correctly after the camera is connected so I don’t understand what’s wrong – and of course, this has worked fine for all the previous QHY cameras.  It may be worth remembering that we’re apparently dealing with a beta version of the camera driver here.

daleghent:
[...]

PR #1246, opened 2024-10-10
Want bit mode selection for QHY cameras (#1245)

daleghent:
Per #1245 , this enhances the QHY native driver by:

* Removing the statically-set 8bpp depth in `m_bpp` variable initialization
* Adding a setup dialog and config path with two radio buttons for selecting 8 or 16 bit output
  - 8bpp output is the default to maintain existing behavior and dark library compatibility
* Camera initialization now always sets up the output bit depth by calling `SetQHYCCDBitsMode(handle, bpp)`

### DIFF
--- a/src/cam_qhy.cpp
+++ b/src/cam_qhy.cpp
@@ -40,6 +40,8 @@
 # include "cam_qhy.h"
 # include "qhyccd.h"
 
+# define CONFIG_PATH_QHY_BPP "/camera/QHY/bpp"
+
 class Camera_QHY : public GuideCamera
 {
     qhyccd_handle *m_camhandle;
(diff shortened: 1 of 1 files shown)


bwdev01:
Hi Dale.  Did you intend to bail out on this change?  I wanted to look at your PR this morning and there’s nothing to see, your branch was deleted.  Did you change your mind about something?

 

Bruce

daleghent:
Hi Bruce, I was testing some additional QHY models I had on hand and realized that fixing this issue may be more fraught than first realized. It's possible we may be looking at an unavoidable breaking change from _some_ but perhaps not _all_ QHY native driver users here. I am testing with a QHY5III462C and QHY183M currently. I feel that I need a wider breadth of test articles to fully map out the issue as it seems the default bit mode can vary between models. I have a QHY5III174M on a telescope at the moment, so I will get that and use it as another data point.

bwdev01:
Ok, that makes sense.  We appreciate the help on this – I was mostly concerned that you might have gotten fouled up with GitHub and inadvertently lost some work.  It might be necessary to wait for QHY to clean up the implementation.  For example, the slow performance in 8-bit mode sounds like a bug and if their ASCOM driver worked correctly in both modes, users could just rely on that.

 

Thanks,

Bruce

PR #1249, opened 2024-10-28
QHY driver improvements

daleghent:
This PR addresses the points raised in #1245 

The new options window presents thusly:

image

Tooltips are provided for each option that explains its purpose and any model-specific relevancy.

### DIFF
--- a/src/cam_qhy.cpp
+++ b/src/cam_qhy.cpp
@@ -1087,7 +1087,7 @@ bool Camera_QHY::SetQhyGain(int gainPercent)
 {
     double gain = m_gainMin + gainPercent * (m_gainMax - m_gainMin) / 100.0;
 
-    if (gain != 0.)
+    if (gain != 0)
         gain = floor(gain / m_gainStep) * m_gainStep;
 
     uint32_t rv = SetQHYCCDParam(m_camhandle, CONTROL_GAIN, gain);
@@ -1120,6 +1120,9 @@ bool Camera_QHY::SetQhyOffset(int offset)
     if (offset != 0)
         offset = floor(offset / m_offsetStep) * m_offsetStep;
 
+    if (offset != m_offset)
+        pConfig->Profile.SetInt(CONFIG_PATH_QHY_OFFSET, offset);
+
     uint32_t rv = SetQHYCCDParam(m_camhandle, CONTROL_OFFSET, (double)offset);
 
     if (rv != QHYCCD_SUCCESS)
@@ -1149,6 +1152,9 @@ bool Camera_QHY::SetQhyUsbTraffic(int usbTraffic)
 
     if (usbTraffic != 0)
         usbTraffic = floor(usbTraffic / m_usbTrafficStep) * m_usbTrafficStep;
+
+    if (usbTraffic != m_usbTraffic)
+        pConfig->Profile.SetInt(CONFIG_PATH_QHY_USBTRAFFIC, usbTraffic);
 
     uint32_t rv = SetQHYCCDParam(m_camhandle, CONTROL_USBTRAFFIC, (double)usbTraffic);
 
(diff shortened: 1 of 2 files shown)


daleghent:
A little editorializing: Release notes ought to mention that it would be a good idea for QHY guide camera users to do a fresh dark library run for their camera. While I took pains with the defaults to ensure that cameras will produce the same output as before this change, that can't be guaranteed on a model-by-model basis. Plus, a fresh dark library is good hygiene anyhow.

src/cam_qhy.cpp
```diff
@@ -556,6 +900,25 @@ bool Camera_QHY::Capture(int duration, usImage& img, int options, const wxRect&
...
+        {
+            m_offset = m_offsetMin;
+        }
+
+        if (m_offset > m_offsetMax)
+        {
+            m_offset = m_offsetMax;
+        }
+
+        double offset = floor(m_offset / m_offsetStep) * m_offsetStep;
```
agalasso, line 914: can we add some protection against `m_offsetStep == 0`, either here or back when we call `GetQHYCCDParamMinMaxStep`?

src/cam_qhy.cpp
```diff
@@ -572,14 +935,29 @@ bool Camera_QHY::Capture(int duration, usImage& img, int options, const wxRect&
...
+        {
+            Debug.Write(wxString::Format("QHY: 2000ms delay needed\n"));
+            WorkerThread::MilliSleep(2000);
+        }
     }
-    if (ret == QHYCCD_READ_DIRECTLY)
+    else
     {
-        // Debug.Write("QHYCCD_READ_DIRECTLY\n");
+        Debug.Write(wxString::Format("QHY: %dms delay needed\n", duration));
```
agalasso, line 959: optional: maybe remove this? may be a little excessive logging this especially for small durations where the logging it self could even take longer than the requested delay

src/cam_qhy.h
```diff
@@ -35,6 +35,9 @@
 #ifndef CAM_QHY_INCLUDED
 #define CAM_QHY_INCLUDED
 
+#define QHYCCD_OFF  0.0
+#define QHYCCD_ON   1.0
+
```
agalasso, line 40: let's move these definitions into cam_qhy.cpp as they do not need to be exposed here as part of the pure interface being defined in this file.

agalasso (commented):
great stuff, thanks!

src/cam_qhy.cpp
daleghent: Definitely. I had that there to follow what was there previously as well as to observe how different camera models did their return codes for the exposure. While neat, it's certainly not needed.

src/cam_qhy.h
daleghent: Done

src/cam_qhy.cpp
```diff
@@ -291,6 +584,36 @@ bool Camera_QHY::Connect(const wxString& camId)
         return CamConnectFailed(_("Failed to get gain range"));
     }
 
+    if (GetQHYCCDParamMinMaxStep(m_camhandle, CONTROL_OFFSET, &m_offsetMin, &m_offsetMax, &m_offsetStep) == QHYCCD_SUCCESS)
```
agalasso, line 587: could we do something like this so if the API call sets m_offsetStep to zero we don't get a divide by zero later on.  Should never happen if the SDK is behaving properly, but prefer not to make that assumption.
```suggestion
    if (GetQHYCCDParamMinMaxStep(m_camhandle, CONTROL_OFFSET, &m_offsetMin, &m_offsetMax, &m_offsetStep) == QHYCCD_SUCCESS && m_offsetStep > 0)
```

agalasso (commented):
just one comment.. if we could resolve that then happy to merge it and get it into the next release

src/cam_qhy.cpp
daleghent: Hey Andy, I made that change (and now guard against a 0 step in several places) and a few others, as well.

daleghent:
I've collected a small list of folks who are interested in testing this new code. I'm sending an email to this evening with instructions. I'll CC you and Bruce and BCC the interested parties.

daleghent:
I've ran tests of the current code overnight and no memory leaks are apparent.

agalasso (approved):
great work, thanks!

PR #1250, opened 2024-11-01
Add cooler control API method

daleghent:
This adds a new JSON-RPC API method, `set_cooler_state`, which takes a single boolean argument to turn the cooler on or off if the connected camera is equipped with one.

Turning the cooler on:
```json
{"method":"set_cooler_state", "params": [ true ], "id":1}
```

Turning it off:
```json
{"method":"set_cooler_state", "params": [ false ], "id":1}
```

This will permit full automated setup of the camera from an API client. Currently, the cooler must be manually turned on via the Advanced Settings > Camera window. This complements the existing `get_cooler_status` method.

### DIFF
--- a/src/event_server.cpp
+++ b/src/event_server.cpp
@@ -2119,7 +2119,7 @@ static void set_cooler_state(JObj& response, const json_value *params)
     if (enable)
     {
         double setpt = pConfig->Profile.GetDouble("/camera/CoolerSetpt", 10.0);
-        if (pCamera->SetCoolerSetpoint(setpt));
+        if (pCamera->SetCoolerSetpoint(setpt))
         {
             response << jrpc_error(1, "failed to set cooler setpoint");
             return;
--- a/src/event_server.cpp
+++ b/src/event_server.cpp
@@ -2089,7 +2089,6 @@ static void get_calibration_data(JObj& response, const json_value *params)
 
 static void set_cooler_state(JObj& response, const json_value *params)
 {
-    bool ret;
     Params p("enabled", params);
     const json_value *val = p.param("enabled");
     bool enable;
@@ -2111,21 +2110,22 @@ static void set_cooler_state(JObj& response, const json_value *params)
         return;
     }
 
+    if (!pCamera->SetCoolerOn(enable))
+    {
+        response << jrpc_error(1, "failed to set cooler state");
+    }
+
     if (enable)
     {
         double setpt = pConfig->Profile.GetDouble("/camera/CoolerSetpt", 10.0);
-        ret = pCamera->SetCoolerSetpoint(setpt);
-        if (!ret)
+        if (!pCamera->SetCoolerSetpoint(setpt));
         {
             response << jrpc_error(1, "failed to set cooler setpoint");
             return;
         }
     }
 
-    if (pCamera->SetCoolerOn(enable))
-        response << jrpc_result(0);
-    else
-        response << jrpc_error(1, "failed to set cooler state");
+    response << jrpc_result(0);
 }
 
 static void get_cooler_status(JObj& response, const json_value *params)
--- a/src/event_server.cpp
+++ b/src/event_server.cpp
@@ -2113,6 +2113,7 @@ static void set_cooler_state(JObj& response, const json_value *params)
     if (pCamera->SetCoolerOn(enable))
     {
         response << jrpc_error(1, "failed to set cooler state");
+        return;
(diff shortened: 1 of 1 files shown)


src/event_server.cpp
```diff
@@ -2087,6 +2087,47 @@ static void get_calibration_data(JObj& response, const json_value *params)
...
+
+    if (!pCamera->HasCooler)
+    {
+        response << jrpc_error(1, "camera lacks a cooler");
+        return;
+    }
+
+    if (!pCamera->SetCoolerOn(enable))
+    {
+        response << jrpc_error(1, "failed to set cooler state");
```
agalasso, line 2115:
```suggestion
        response << jrpc_error(1, "failed to set cooler state");
        return;
```

src/event_server.cpp
```diff
@@ -2087,6 +2087,47 @@ static void get_calibration_data(JObj& response, const json_value *params)
...
+        return;
+    }
+
+    if (!pCamera->HasCooler)
+    {
+        response << jrpc_error(1, "camera lacks a cooler");
+        return;
+    }
+
+    if (!pCamera->SetCoolerOn(enable))
```
agalasso, line 2113: unfortunately the `Camera` and `Guider` APIs use the odd convention--predating my involvement in the project--returning `false` on success (!).  I'd love to change that at some point, but have never gotten around to it.
```suggestion
    bool error = pCamera->SetCoolerOn(enable);
    if (error)
```

src/event_server.cpp
```diff
@@ -2087,6 +2087,47 @@ static void get_calibration_data(JObj& response, const json_value *params)
...
+    {
+        response << jrpc_error(1, "failed to set cooler state");
+    }
+
+    if (enable)
+    {
+        double setpt = pConfig->Profile.GetDouble("/camera/CoolerSetpt", 10.0);
+        if (!pCamera->SetCoolerSetpoint(setpt));
```
agalasso, line 2121: ditto here (SetCoolerSetpoint returns false on success 🤢 )

src/event_server.cpp
daleghent: Would correcting this be in-scope for this change? I don't really think that anything else actually pays attention to the return values of the cooler funcs.

agalasso: You're right, other callers currently ignore the return value of SetCoolerOn.
I don't think that changing _just_ SetCoolerOn to have a different convention from the other APIs would be a good idea --  as it is now it may be an odd convention, but at least it is consistent through all the camera and guider APIs.  Having some APIs return false and others return true for success will make the code more fragile IMO.  My preference would be to not change the convention in this pr.  We can do a followup change at some point to fix it across all the APIs (or we can just leave it alone.)

daleghent: Alright. Because I was testing this new API method with a cooled QHY camera, I "fixed" the return boolean in the QHY driver improvements PR that was merged today, so I'll need to undo those and then reverse the logic in this one to match the, erm, convention ;) I have some feedback on camera functions from some folks so am working on a follow-up QHY PR right now, so I'll un-correct the return boolean in `SetCoolerOn()` and `SetCoolerSetpoint()` functions there as a part of that.

src/event_server.cpp
bwdev01: I agree. As annoying as the current convention is, I've gotten used to it and it is at least consistent.  I think a partial, mixed convention would make life harder and contribute to programming errors.

src/event_server.cpp
daleghent: function is `void` and none of the existing `void`-typed method handler funcs explicitly `return` at the end. Should this be expanded to cover the preexisting functions as well, or just this one?

agalasso: sorry, I should have been more clear on why I made the suggestion. If the SetCoolerOn call fails and we generate an error response (line 2115), then this method is done and needs to return at this point and not keep going -- if it kept going it would append additional failure or success info to the response which would be wrong.

daleghent: Oh! Right. I couldn't quite tell the line in this late hour. Fixed.

agalasso (approved):

agalasso:
@daleghent thanks for the pr! merging it now.
when you get a chance could you please update the documentation on the wiki page here https://github.com/OpenPHDGuiding/phd2/wiki/EventMonitoring

PR #1251, opened 2024-11-03
QHY driver followups

daleghent:
This is a follow-up that contains changes based on feedback from some testers. Some items are clarified in better ways, and the boolean return values for the cooler functions are corrected to follow PHD2 convention.


agalasso (approved):
great!

PR #1259, opened 2024-11-22
Final QHY improvements

daleghent:
This is the final installment of my QHY native driver project. This has gone through some intensive testing with some users so it lagged a little.

The two main themes of this are:

1. Detecting if 16bit mode is even possible and reacting appropriately. This is needed for the original QHY5II (non-L) camera, apparently, which ran in 8bit mode only
2. Always set the exposure duration before initiation of camera exposure. This resets the camera's internal rolling loop of sensor readouts and resets so that the resulting frame is aligned as close as possible to the time which PHD2 initiates the exposure and when it is available. There has been some talk in the past about this issue with QHY cameras and how it can negatively affect the guiding performance of some mounts. There's no harm in making this change, so it is as good opportunity to do so.

### DIFF
--- a/src/cam_qhy.cpp
+++ b/src/cam_qhy.cpp
@@ -86,7 +86,6 @@ class Camera_QHY : public GuideCamera
     double m_devicePixelSize;
     unsigned char *RawBuffer;
     wxSize m_maxSize;
-    int m_curExposure;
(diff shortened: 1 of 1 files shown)


agalasso (approved):
nice work, thanks!

PR #1260, opened 2024-11-27
Image scale change improvements

bwdev01:
Changes to pixel size, focal length, or binning currently trigger clearing of calibration.  This is unnecessary for minor changes in focal length or pixel size.  Also, the current floating point equality check always fails for reported pixel sizes on SX cameras because of approximation.

### DIFF
--- a/src/advanced_dialog.cpp
+++ b/src/advanced_dialog.cpp
@@ -592,6 +592,21 @@ double AdvancedDialog::DetermineGuideSpeed()
     }
     return sidRate;
 }
+
+// Floating point equality comparisons can be a problem due to rounding and trivial inequalities.
+// PercentChange can be used to see if a change is worth reacting to
+double AdvancedDialog::PercentChange(double oldVal, double newVal)
+{
+    double chg;
+    if (oldVal > 0)
+    {
+        chg = fabs(1.0 - newVal / oldVal);
+    }
+    else
+        chg = fabs(newVal);
+    return 100.0 * chg;
+}
+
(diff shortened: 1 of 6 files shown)


src/advanced_dialog.cpp
```diff
@@ -592,6 +592,21 @@ double AdvancedDialog::DetermineGuideSpeed()
...
+// PercentChange can be used to see if a change is worth reacting to
+double AdvancedDialog::PercentChange(double oldVal, double newVal)
+{
+    double chg;
+    if (oldVal > 0)
+    {
+        chg = fabs(1.0 - newVal / oldVal);
+    }
+    else
+        chg = fabs(newVal);
```
agalasso, line 606: hmm, the negative case (oldVal < 0) seems odd -- probably won't happen in this pr but could be an unexpected result if somebody else wants to use this method. how about something like this:
```
    if (fabs(oldVal) < 0.0001)
        return 100. * newVal; // not meaningful, but avoids divide by zero
    return 100. * fabs(1. - newVal / oldVal);
```

bwdev01: Good idea.  This was originally a private static function in camera.cpp but when I promoted it to advanced_dialog, I should have paid more attention to the general use cases.

bwdev01:
Good point.  This was originally a private static function in camera.cpp but I later promoted it to advanced_dialog.cpp and neglected to think about the other use-cases.  I just used your suggested change so I went ahead and merged the PR without asking for another review.

 

Thanks.

PR #1263, opened 2024-12-09
Avoid PPEC crash during dithering ops

bwdev01:
This is the most restrictive approach because it only applies during dithering.  The only identified (rare) field problems occurred during dithering and settling.  Generalizing the protection beyond dithering would take more work.

### DIFF
--- a/contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
+++ b/contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
@@ -278,11 +278,13 @@ double GaussianProcessGuider::result(double input, double SNR, double time_step,
         {
             deduceResult(time_step); // just pretend we would do dark guiding...
         }
-        catch (std::runtime_error err)
+        catch (const std::runtime_error err)
         {
             reset();
-            std::string outStr = "PPEC: Model reset after exception: " + std::string(err.what());
-            GPDebug->Log(outStr.c_str());
+            std::ostringstream message;
+            message << "PPEC: Model reset after exception: " << err.what();
+            GPDebug->Write(message.str().c_str());
+
             return parameters.control_gain_ * input;
         }
 
--- a/contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
+++ b/contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
@@ -60,21 +60,11 @@
 
 #define HYSTERESIS 0.1 // for the hybrid mode
 
-GaussianProcessGuider::GaussianProcessGuider(guide_parameters parameters) :
-    start_time_(std::chrono::system_clock::now()),
-    last_time_(std::chrono::system_clock::now()),
-    control_signal_(0),
-    prediction_(0),
-    last_prediction_end_(0),
-    dither_steps_(0),
-    dithering_active_(false),
-    dither_offset_(0.0),
-    circular_buffer_data_(CIRCULAR_BUFFER_SIZE),
-    covariance_function_(),
-    output_covariance_function_(),
-    gp_(covariance_function_),
-    learning_rate_(DEFAULT_LEARNING_RATE),
-    parameters(parameters)
+GaussianProcessGuider::GaussianProcessGuider(guide_parameters parameters)
+    : start_time_(std::chrono::system_clock::now()), last_time_(std::chrono::system_clock::now()), control_signal_(0),
+      prediction_(0), last_prediction_end_(0), dither_steps_(0), dithering_active_(false), dither_offset_(0.0),
+      circular_buffer_data_(CIRCULAR_BUFFER_SIZE), covariance_function_(), output_covariance_function_(),
+      gp_(covariance_function_), learning_rate_(DEFAULT_LEARNING_RATE), parameters(parameters)
 {
     circular_buffer_data_.push_front(data_point()); // add first point
     circular_buffer_data_[0].control = 0; // set first control to zero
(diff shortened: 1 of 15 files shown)


src/star.cpp
```diff
@@ -1031,8 +1031,9 @@ bool GuideStar::AutoFind(const usImage& image, int extraEdgeAllowance, int searc
             // We're repeating the find, so we're vulnerable to hot pixels and creation of unwanted duplicates
             if (tmp.WasFound() && tmp.SNR >= minSNR)
             {
-                bool duplicate = std::find_if(foundStars.begin(), foundStars.end(), [&tmp](const GuideStar& other)
-                                              { return CloseToReference(tmp, other); }) != foundStars.end();
+                bool duplicate =
+                    std::find_if(foundStars.begin(), foundStars.end(),
+                                 [&tmp](const GuideStar& other) { return CloseToReference(tmp, other); }) != foundStars.end();
```
agalasso, line 1036: this change seems to be unrelated to the pr ... let's revert it and get star.cpp out of the pr

contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
```diff
@@ -274,7 +274,17 @@ double GaussianProcessGuider::result(double input, double SNR, double time_step,
...
-        deduceResult(time_step); // just pretend we would do dark guiding...
+        try
+        {
+            deduceResult(time_step); // just pretend we would do dark guiding...
+        }
+        catch (std::runtime_error err)
+        {
+            reset();
+            std::string outStr = "PPEC: Model reset after exception: " + std::string(err.what());
+            GPDebug->Log(outStr.c_str());
```
agalasso, line 285: Log takes a printf-style format string so this could be a little dangerous (I'm thinking of the case where outStr contains a `%` character somewhere)
Here's is a safer way that also gets rid of the unneeded temporary string construction:
```suggestion
            GPDebug->Log("PPEC: Model reset after exception: %s", err.what());
```

contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
```diff
@@ -274,7 +274,17 @@ double GaussianProcessGuider::result(double input, double SNR, double time_step,
         {
             dithering_active_ = false;
         }
-        deduceResult(time_step); // just pretend we would do dark guiding...
+        try
+        {
+            deduceResult(time_step); // just pretend we would do dark guiding...
+        }
+        catch (std::runtime_error err)
```
agalasso, line 281: minor: can think of the catch args as like method arguments, and it is a good practice to pass objects by const reference where we want reference semantics (pass by reference) and do not need copy semantics (pass by value):
```suggestion
        catch (const std::runtime_error& err)
```

contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
```diff
@@ -689,38 +699,53 @@ Eigen::MatrixXd GaussianProcessGuider::regularize_dataset(const Eigen::VectorXd&
...
+
+                    gear_error_sum = 0.0;
+                    variance_sum = 0.0;
+                    ++j;
+                }
+            }
+        }
+        catch (std::runtime_error err)
+        {
+            throw err;
```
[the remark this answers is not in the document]
bwdev01: It won't compile without it.  The error message is that at least one handler must be specified for the try block in the function.

src/star.cpp
bwdev01: This is the same old problem with clang and star.cpp.   This has reverted the clang change without me ever opening or touching that file.  Something is seriously messed up with this.

contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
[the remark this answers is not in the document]
agalasso: right, both the try and catch both should be removed

src/star.cpp
agalasso: we'll need to find out what is changing that file in your local environment, and also be sure not to commit the file -- you can run `git restore src/star.cpp` before you commit it if you notice that it has been changed unintentionally.
Now that the "bad" version has been committed into the branch you can get the old version back with
```
git restore -s origin/master -- src/star.cpp
```
(or use any other version you want in place of `origin/master`)
Then `git add` `git commit` and `git push` to get the branch fixed up on github (origin).

bwdev01:
Ok, you’ve lost me here.  Are you proposing that the throw shouldn’t be enclosed in a try/catch block?  The throw I’m forcing is at line 719 in that source.  Even if C++ allows a throw outside of a try/catch block, what happens?  The goal here is to trigger an exception before the crash occurs and have that exception caught in a function two levels down in the call stack at a point where recovery is well-defined.

bwdev01:
Regarding the log output for the throw message.  Doing it the way you suggested results in this:

 

19:34:33.706 09.355 26228 PPEC: Model reset after exception: 湉敤⁸癯牥爭湵椠⁮敲畧慬楲敺摟瑡獡瑥ﴀ﷽ý0

 

I started out trying to use the %s format specifier and chased my tail trying to get anything usable – that’s how I ended up with my first try.  If the %s format is expecting a wxString, we’re out of luck because the gp* functions don’t use wxWidgets.  Any other approaches come to mind?

agalasso:
> Regarding the log output for the throw message. Doing it the way you suggested results in this

@bwdev01 GPDebug::Log is not working correctly for `%s` arguments -- a latent bug this PR is now triggering it as it is the first use of `%s` in GPDebug.    I'm looking into it now...

agalasso:
@bwdev01 if #1264 looks ok to you, we can merge that to master so you can make use of it in this pr, something like this:
```
    std::ostringstream message;
    message << "PPEC: Model reset after exception: " << err.what();
    GPDebug->Write(message.str().c_str());
```

contributions/MPI_IS_gaussian_process/src/gaussian_process_guider.cpp
```diff
@@ -274,7 +274,19 @@ double GaussianProcessGuider::result(double input, double SNR, double time_step,
         {
             dithering_active_ = false;
         }
-        deduceResult(time_step); // just pretend we would do dark guiding...
+        try
+        {
+            deduceResult(time_step); // just pretend we would do dark guiding...
+        }
+        catch (const std::runtime_error err)
```
agalasso, line 281: we can use pass-by-reference semantics here rather than pass-by-value:
```suggestion
        catch (const std::runtime_error& err)
```

agalasso (commented):
looks good, one minor suggestion

agalasso (approved):
nice work

PR #1264, opened 2024-12-14
Workaround for GPDebug::Log not working with format %s

agalasso:
Add a Write method to GPDebug::Log for logging an already formatted C-style string. This is a workaround for GPDebug::Log not being able to handle a format string with %s.


bwdev01 (approved):
This will work, let's merge it and I will use it for the other PR.

PR #1267, opened 2024-12-16
Predictive PEC: use the monotonic clock rather than the system wall clock

agalasso:
GP guider needs accurate time interval measurements, and is not interested in the wall clock.  std::chrono::steady_clock is a better choice than system_clock.

Testing:
 - gp guider unit tests pass


bwdev01 (approved):
Looks good, this should eliminate an obvious weakness

PR #1278, opened 2025-01-02
PlayerOne cameras: update to SDK version V3.7.1

agalasso:
* Merge prior work on PlayerOne cameras with Ethan's
 * format source files with clang-format
 * tested with vendor's camera on Mac and Windows

PR #1276, opened 2024-12-31
Native support for Player One cameras

EthanChappel:
This pull request is for adding native support for cameras from Player One Astronomy, which I've seen brought up in #1238 in August but haven't seen signs of movement on since. Based on the code for supporting ZWO cameras in PHD2 since the SDKs for both manufacturers are similar.

**Tested cameras**
Xena-M
Uranus-C

**Tested Platforms:**
Windows 11 x64
Linux ARM32 (Raspberry Pi 2)
Linux ARM64 (Raspberry Pi 5)

Camera controls for exposure, gain, bit depth, and binning work as expected on tested cameras and platforms. Can test cooling control soon with a Uranus-C Pro or Poseiden-M Pro.

Builds successfully on x86 and x64 Linux but not tested yet. Currently facing issues building PHD2 on macOS Sequoia on an Intel Mac, so may need help testing if I am unable to get it to succeed.

image

PR #1282, opened 2025-01-04
Windows build: improvement to DLL copying

agalasso:
Copy the Windows DLLs with a single call to `cmake -E` rather than one call per DLL. This slightly speeds up the build for a better experience in iterative development on Windows.

Also moves the DLL copying to POST_BUILD rather then PRE_BUILD as there's no point in copying the DLLs if there's going to be a compile error during iterative development.


agalasso:
re: clang-format check failure -- it's unrelated to this pr and got introduced by #1276 and is fixed by #1278

PR #1286, opened 2025-01-06
New class for camera selector button & use in Profile Wizard

EthanChappel:
Moves much of the code for the camera selector button out of the gear dialog into a new class. Does not change anything from the user's perspective, but would like to use this new class to add the button to Profile Wizard so we no longer have to hope the right camera is selected if multiple using the same driver are attached to the PC. Most of the work is done there too, but I plan to finish and submit after this pull request is finalized and merged.

Tested on Windows 11 with one ZWO and two Player One cameras.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -354,6 +354,8 @@ set(phd2_SRC
   ${phd_src_dir}/calstep_dialog.h
   ${phd_src_dir}/camcal_import_dialog.cpp
   ${phd_src_dir}/camcal_import_dialog.h
+  ${phd_src_dir}/camera_selector_button.cpp
+  ${phd_src_dir}/camera_selector_button.h
   ${phd_src_dir}/circbuf.h
 
   ${phd_src_dir}/comet_tool.cpp
--- a/src/camera_selector_button.cpp
+++ b/src/camera_selector_button.cpp
@@ -40,7 +40,7 @@ wxDEFINE_EVENT(SELECT_CAMERA_EVENT, wxCommandEvent);
 wxBEGIN_EVENT_TABLE(CameraSelectorButton, wxBitmapButton)
 	EVT_BUTTON(GEAR_BUTTON_SELECT_CAMERA, CameraSelectorButton::OnButtonSelectCamera)
     EVT_MENU_RANGE(MENU_SELECT_CAMERA_BEGIN, MENU_SELECT_CAMERA_END, CameraSelectorButton::OnMenuSelectCamera)
-wxEND_EVENT_TABLE()
+wxEND_EVENT_TABLE();
 // clang-format on
 
 CameraSelectorButton::CameraSelectorButton(wxWindow *pParent, wxWindowID id)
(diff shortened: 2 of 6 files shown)


bwdev01 (commented):
What is the point of all this?

EthanChappel:
This is why I am hoping to have the new class reviewed and merged first instead of lumping it all in one PR. Separating the button into a new class introduced issues in the gear dialog that needed to be resolved first (i.e. selected camera model not changing when the driver changed, so it could try connecting to a non-existent Player One ASI120MM or ZWO Ceres 462M). I believe they are all ironed out.

Working on the profile wizard part on [profile-wizard-cam-select-btn](https://github.com/EthanChappel/phd2/tree/profile-wizard-cam-select-btn). To briefly summarize, it seems to work as intended and not affect other profiles so far. Still testing various scenarios and may have to make further changes if any are made to the new button class during this review.

bwdev01:
In your newly created class, it seems to me that any reference to "pConfig->Profile" is a problem if the class is going to be used by the Profile Wizard.  That reference is to the already loaded profile not to the profile you are trying to build.  There is no "new profile" at this point.  It's not a trivial problem.  Andy and I are about to do another dev release around the end of the week, so I'd like to see this wizard-related stuff put on the shelf until we can get through with that and until a better approach can be worked out.  FWIW, I think it will generally work better if you make contact ahead of time and discuss proposed changes before plunging into coding.  For example, if I had known this was in the pipeline, I wouldn't have just spent the time to rework the Help content for using multiple cameras from the same manufacturer.

EthanChappel:
I am back to working on this.

Since the discussion is revolving around the end goal of using the button in the Profile Wizard, I've merged the work for that into this branch rather than opening a separate PR after this one.

The button works for me as-is in Profile Wizard since the AutoTempProfile used by it becomes pConfig while the wizard is open. The [AutoTempProfile constructor](https://github.com/OpenPHDGuiding/phd2/blob/0927de6c8943fae7161457008b989bf72a05c638/src/phdconfig.cpp#L901) has an init parameter which determines whether the temp profile is set to pConfig. As far as I can tell, this is always true in practice.

I can move the code for setting the camera model from the button's class to the parent dialogs if desired. This is how it was originally being implemented it until realizing it wasn't necessary to work, so for now I believe the end result would be duplicate code.

bwdev01:
I’ll try to take a look at this soon, maybe over the weekend.

 

Bruce

bwdev01:
I just cloned your github branch and I can't get the project to build.  I've confirmed that the source tree from the PHD2 repo builds on this system so the dev environment is correct.  The build for the clone of your branch is failing on the ogma camera sdk code which is a recent addition to the master branch.  Could you please update your branch and confirm that it will build correctly from the run_cmake.bat step.  Thanks.

bwdev01:
Hi Ethan.  Thanks for giving us a shove on this item.  I just committed changes needed to provide the capability in the ProfileWizard but it takes a different approach than what you pursued.  I went with a lighter-weight approach that didn't rely on the user clicking the appropriate buttons to fully specify the device id, something I think is more appropriate for a wizard interface and new users.  Cheers.

PR #1288, opened 2025-01-10
Fix crash with Player One cameras on Linux

EthanChappel:
Attempting to select or connect to Player One cameras on Linux has crashed the program since #1278.  Renaming the `POACamera` class to something else like `PlayerOneCamera` fixes the issue, since the Player One SDK may already uses the name.

Tested on x64, Arm64 Debian 12 and Windows 11.

### DIFF
--- a/src/cam_playerone.cpp
+++ b/src/cam_playerone.cpp
@@ -55,7 +55,7 @@ enum CaptureMode
     CM_VIDEO,
 };
 
-class POACamera : public GuideCamera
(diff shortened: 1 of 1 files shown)


agalasso:
@EthanChappel thanks for tracking that down. Interesting that I never saw the issue on my Linux, Mac or Windows systems so hopefully this really is the root cause.  Even if it is not the root cause, the change is harmless and worth taking in case it really is the root cause.

PR #1292, opened 2025-01-25
Andy/planetary

bwdev01:
This fixes a long-standing mistake in creating the Guider window.  It doesn't cause a problem until a second Guider instance is created, at which point the second window instance always has a size of (20,20).  The desired size parameters were always being passed but were being ignored.


agalasso:
@bwdev01 The pr includes all the changes from the planetary branch, surely not the intent, right?   I am going to close the pr so it does not get merged inadvertently.  I'll send you an email and we can work out how to create a new pr with just the changes you want.

PR #1304, opened 2025-04-15
Profile wiz

bwdev01:
Adds ability in Profile Wizard to specify a particular camera device within a camera/manufacturer type.  This is akin to the 'branch' icon function in the Connect Equipment dialog but doesn't rely on the user clicking on a button.  The behavior looks like this for any camera type that returns true for CanSelectCamera():
1.  Camera not physically connected - no change from previous behavior.
2.  Single camera connected - the camera device id is obtained automatically, displayed in a label box below the camera selection drop-down, and is included in the camera/hash property of the profile once the profile is completely specified.
3. More than 1 camera connected - pop-up menu is automatically displayed and user selection is then handled as in item 2.
4. Changes in the camera selection drop-down automatically clear earlier device-related properties and repeat the process described above.
None of this applies to ASCOM or Indi drivers which return false for CanSelectCamera.  This implementation is entirely different from the earlier PR on the same subject.  This PR has been field-tested by Brian Valente using multiple combinations of guide cameras.


src/profile_wizard.cpp
```diff
@@ -875,26 +902,37 @@ struct AutoConnectCamera
...
         {
-            wxMessageBox(_("PHD2 could not connect to the camera so you may want to deal with that later. "
+            wxString preamble = (m_camera->CanSelectCamera() && parent->NumCamerasFound() == 0)
+                ? _("No cameras of that type were found")
+                : _("PHD2 cound not connect to the camera");
+
+            wxMessageBox(preamble +
+                         _(" so you may want to deal with that later. "
                            "In the meantime, you can just enter the pixel-size manually along with the "
                            "focal length and binning levels."));
```
agalasso, line 931: It will be better for translators if you make two separate strings rather than a concatenation of preamble + rest. When translating to another language, it may be more natural to express it some other way than preamble + " so ...".

src/profile_wizard.cpp
```diff
@@ -914,13 +952,20 @@ struct AutoConnectCamera
...
         cam->SetBinning(val);
 }
 
+static wxString CameraSelectionKey(const wxString& camName)
+{
+    std::hash<std::string> hash_fn;
+    std::string name(camName.c_str());
+    return wxString::Format("/cam_hash/%lx/whichCamera", (unsigned long) hash_fn(name));
+}
+
```
agalasso, line 968: Let's not have two copies of this code.  On way to fix it would be to move the original code into a public static method of GearDialog and call it from the places it is needed.

src/profile_wizard.cpp
```diff
@@ -972,6 +1017,12 @@ void ProfileWizard::WrapUp()
     if (ImageScale < 2.0)
         pConfig->Profile.SetBoolean("/guider/onestar/MassChangeThresholdEnabled", false);
     pConfig->Profile.SetInt("/camera/SaturationADU", 0); // Default will be updated with first auto-find to reflect bpp
+    if (m_camDeviceId != GuideCamera::DEFAULT_CAMERA_ID)
+    {
+        wxString key = CameraSelectionKey(m_SelectedCamera);
+        const wxString& id = m_camDeviceId;
+        pConfig->Profile.SetString(key, id);
```
agalasso, line 1024: nit:
```suggestion
        pConfig->Profile.SetString(key, m_camDeviceId);
```

agalasso (commented):
nice work!

bwdev01:
Thanks, Andy, I’ll make those changes.

agalasso (approved):

PR #1318, opened 2025-06-28
Fix ToupTek/OGMA sensor temperature

daleghent:
The `GetCoolerStatus()` function for ToupTek and OGMA cameras did not set the `temperature` pointer with the cooler temperature readout, causing PHD2 to display garbage of an uninitialized variable. This PR fixes that so that the temperature is fetched and thus properly displayed. Although also ToupTek-based, the Malincam and Altair drivers leave these functions unimplemented and I felt that adding them would be outside the scope of this PR, plus I don't have any such devices to test with to ensure proper function. I tested this fix with a ToupTek ATR585M.

Before:
image

After:
image


agalasso (commented):
Thanks for the PR!

PR #1319, opened 2025-07-06
Touptek scaling

john-stone-ics:
After extensive conversation with ToupTek tech support it has been determined that when in raw 16-bit mode different camera's return different ranges of ADU values.

The raw bit depth is returned by 

Toupcam_get_RawFormat(m_cam.m_h, &fourcc, &bpp)

For example my GPM462M returns ADU counts in the range [0 - **4094**  ] ; which is (2^12 - 2) and bpp = 12

My G3M678M return ADU counts in the range [0 - **65520**] which is [0 - **4095**  ] << 4; or (2^12-1) << 4 and bpp = 16

Touptek has confirmed that their 14-bit cameras (like the ATR533M) return values in the range [0 - **16382**]; which is (2^14 - 2) and bpp = 14

This PR simply scales these output ranges to the full [0 - 65535] expected by the reset of PhD2 code base.

### DIFF
--- a/src/cam_touptek.cpp
+++ b/src/cam_touptek.cpp
@@ -142,7 +142,7 @@ struct ToupCam
             sz->x = width;
             sz->y = height;
 
-            if (m_bpp == 16 && false)
+            if (m_bpp == 16)
             {
                 // There are 12-bit Touptek cameras (like the GPM462M) that return values in the range [0-4094] when in 16-bit mode. 
                 // We scale the 12-bit input range [0-4094] to the 16-bit output range [0-65535].  
--- a/src/cam_touptek.cpp
+++ b/src/cam_touptek.cpp
@@ -52,6 +52,7 @@ struct ToupCam
     void *m_buffer;
     void *m_tmpbuf;
     wxByte m_bpp; // bits per pixel: 8 or 16
+    wxByte m_bppRaw; // returned by Toupcam_get_RawFormat
     bool m_isColor;
     bool m_hasGuideOutput;
     double m_devicePixelSize;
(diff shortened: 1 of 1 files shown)


john-stone-ics:
agalasso,

I haven't had enough clear sky time to find all the problems but using my GPM462M I found it constantly choose saturated stars when clicking on the 'Star' icon on the toolbar.

I also found that my Max ADU setting in the Camera tab of the Advanced dialog kept being changed by to 65535 whenever I clicked that 'Star' icon.

I then found the section of code that was over-writing my saturation threshold and modified it to stop changing it, but then I discovered that saturated stars were still being selected.

I further found that in the star detection code where it subtracted the pedestal before running the star detector that my pedestal was equal to my ADU so the subtraction produced a value of zero.   Modifying the code to ignore the pedestal finally let me camera work correctly.   

I didn't debug further to find out why this was happening.  Instead I decided to scale the image to the correct [0-65535] range in order get the camera working with the rest of the PhD2 code base.   It is my assumption that everything will now work correctly because if it didn't then you would have a lot more bug reports filed, but as I said I haven't had clear sky time to actually verify this myself.

agalasso:
> constantly choose saturated stars when clicking on the 'Star' icon on the toolbar

Would you mind posting a sample image that exhibits the problem (from stock unmodified PHD2) -- I am interested to see where the star finder got fouled up since again Touptek are not the only cameras that do not scale the ADU range to full 16 bits.  (File => Save Image from the menu then attach the image or post a link; including the corresponding Debug Log would be helpful as well)

> my Max ADU setting in the Camera tab of the Advanced dialog kept being changed back to 65535 after setting it to 4094 whenever I clicked that 'Star' icon

Do you happen to have a phd2 debug log from a session where this happens -- if we can reproduce the issue that is definitely something we'll want to fix (independent of this PR) @bwdev01 any idea how this could be happening?

bwdev01:
I think you're missing a lot of context here, things that have been covered in the support forum.  There is no mystery about what is happening on the PHD2 side of things - no code mysteries to be tracked down and resolved.  I added a PR in the dev7 release to help side-step a problem I was seeing all the time on the support forum.  Namely, users were frequently running with a saturation ADU of 255 when the camera was operating in 16-bit mode.  The PR is here: https://github.com/OpenPHDGuiding/phd2/pull/1269
To allow for 16-bit cameras that have 12-bit ADC's and don't re-scale properly,  the PR code will tolerate a saturation level of 4095 - that value will be left alone.  The OP is complaining that his camera saturates at something like 4092, but we know that will make no practical difference in terms of auto-selecting real stars.  So setting a value of 4095 would be the sensible thing to do in order to move ahead and figure out what's actually causing the problem.  In fact, the OP has multiple options that I gave him, none of which require any PHD2 changes:
     Set the saturation adu level to 4095 and work through the wonky gain adjustments on the camera to get usable guide stars (refer to Cloudy Nights on that topic)
    Disable saturation-by-adu and use saturation-by-star-profile (Advanced Settings)
    Revert to the 2.6.13dev5 release of PHD2 until the camera driver gets fixed. That release does not re-adjust ADU settings.
    Use the ASCOM driver for the camera which evidently scales the data sensibly

Since the OP has ignored repeated requests to upload log files, we have almost no visibility.  I don't even know that we're talking about live conditions under a suitable nighttime sky.  I think there's at least a fair chance the problem lies with other camera settings such as gain.  Please follow this link and consider these possibilities:
https://www.cloudynights.com/topic/967761-touptek-gpm462m-guide-camera-gain-psa/

You will notice the original complaint in the CN thread is that all the stars were saturated, fuzzy blobs, just like the screen-shot that appeared in the early part of the PHD2 forum thread.  The side-complaint about the pedestal in the code tells me the OP doesn't understand how dark libraries are implemented, it's a bogus issue.  He can build a better dark library, one with the appropriate gain settings, or he can simply click on the UI button to disable use of dark libraries.  Overall, I think making code changes based only on what we've been told in this thread is unwarranted.

agalasso:
@john-stone-ics I got some more info from @bwdev01 and from reading the forum threads.

The assumption behind this PR that "the full [0 - 65535] expected by the reset of PhD2 code base" is not the intended behavior of PHD2.  The intent is to handle whatever range of ADU values the camera returns.  Ensuring that all cameras use a full 16-bit range is not a sustainable approach since the camera models and drivers are a continually moving target and we prefer not to get into the business of applying scaling on a per-camera basis. It's better in the long term to keep phd2 flexible about accommodating various camera ADU ranges without assumptions.

Any case where an assumption about camera ADU range has crept in to the code is a bug that we will fix. One example of this is the star saturation max ADU threshold not being allowed below 4095 (and quietly overriding the value entered in the UI). We are going to fix that in a separate PR.

Once that bug is fixed, you will be able to set a saturation threshold of 4092 (or whatever) and the star finder should work as expected with your camera.  If that is not the case, please provide us a sample camera frame and we will look into it further.

PR #1321, opened 2025-07-12
Remove auto-adjustment of Saturation-ADU

bwdev01:
Modifies a PR done in dev5 that auto-adjusted saturation-adu level using the bits-per-pixel property of the driver.  This didn't work correctly for drivers that don't scale 10, 12, or 14-bit ADC output to the 16-bit data range.


agalasso (approved):
👍

PR #1326, opened 2025-08-11
ZWO ASI cameras: update to SDK version 1.38

agalasso:
ZWO ASI cameras: update to SDK version 1.38


bwdev01 (approved):
Can you explain how this is tied to 64-bit support?  Is the ZWO SDK inherently 64-bit now?

agalasso:
> Can you explain how this is tied to 64-bit support? Is the ZWO SDK inherently 64-bit now?

Not directly related to 64-bit support, other than that the SDK unpacker script has been updated to no longer ignore the the 64-bit Windows version of the SDK libs when unpacking the zip archive. The zip archive provided by ZWO contains both 32-bit and 64-bit windows libraries, mac libraries, and linux libraries. The script now drops both 32 and 64-bit Windows lib+dll versions into the source tree. The following pr introducing 64-bit windows build capability locates the 32 bit or 64-bit Windows SDK libs as needed.

Other than that change to the unpacker script, this is a standard SDK upgrade like we always do.

PR #1333, opened 2025-08-16
Formatting changes from clang-format

agalasso:
Formatting changes from clang-format


bwdev01 (approved):
Not sure why it's such a bad idea to have "unnecessary parens" in expressions when they serve to re-emphasize the operator precedence.  Is this a Clang "thing"?

bwdev01:
Ok, understood, it's a small point.  But I always find it easier to parse conditional assignments when there are extra parens, especially when the expressions are long or complicated.  Not something worth worrying about now though.

PR #1334, opened 2025-08-16
Rename FullSize to FrameSize

agalasso:
# Based on PR #1333

## Chain of upstream PRs as of 2025-08-16

* PR #1330:
  `master` ← `andy/rename-method`

  * PR #1331:
    `andy/rename-method` ← `andy/spelling-fixes`

    * PR #1332:
      `andy/spelling-fixes` ← `andy/tag-text-for-translation`

      * PR #1333:
        `andy/tag-text-for-translation` ← `andy/clang-format-fixes`

        * **PR #1334 (THIS ONE)**:
          `andy/clang-format-fixes` ← `andy/rename-FullSize-to-FrameSize`

Rename FullSize to FrameSize

FrameSize is a better name since it represents the guide frame image
size which may or not be the same size as the camera's full frame size.

There is one camera (Altair) where it is already the case that the frame size
may be less than the camera sensor size.  We plan on extending this to
other cameras so that it is possible to limit the frame size to be less than
the full camera frame size.


bwdev01 (approved):
So, if you do a search on 'FullSize' across the source base, do you now get zero hits?

agalasso:
> So, if you do a search on 'FullSize' across the source base, do you now get zero hits?

🎯  yes!

PR #1341, opened 2025-08-18
Minor refactor - add a camera Connect wrapper method

agalasso:
Minor refactor - add a camera Connect wrapper method

Add a wrapper method to call Camera::Connect.  This will allow us to
put code common to all camera subclass Connect methods in ConectCamera
rather than repeating it in all camera subclasses' Connect methods.

### DIFF
--- a/src/camera.cpp
+++ b/src/camera.cpp
@@ -655,6 +655,9 @@ GuideCamera *GuideCamera::Factory(const wxString& choice)
     return pReturn;
 }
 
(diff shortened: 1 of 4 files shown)


agalasso:
> Why do we need a static method for this, aren't clients going to have an instance of the camera in any case? Just curious

Good point.  It could just as well have been a non-static member function declared as `final` -- the significant thing is we do not want subclasses overriding it.

I'd be totally fine switching it to a final + non-static member function if you think that would be cleaner (or less unexpected)?

PR #1350, opened 2025-09-08
update Touptek lib to version 59.29465.20250907

touptek:


agalasso (approved):
Thanks for the update!

@touptek  FYI I have started producing 64-bit Windows builds of phd2 so if you want to start including a 64-bin Windows DLL we can enable toupcam cameras in the 64-bit Windows builds.

PR #1356, opened 2025-10-16
 OGMA support is Windows-only and allow OPENSOURCE_ONLY=0 deb file packaging on Linux

d33psky:
Explain to cmake that OGMA support is Windows-only
Allow bundled binary-only libraries with OPENSOURCE_ONLY=0 on Linux


pchev:
Thank you Hans to fix that.
No problem for me, I always use OPENSOURCE_ONLY=1 and the important point is dh_auto_configure is set with this value.

PR #1360, opened 2025-10-23
Fix missing header includes causing compilation errors in Gaussian Process module

darekp:
This pull request adds missing standard library header includes to fix compilation issues in the Gaussian Process code and related tests:

gaussian_process.cpp: Added #include <cassert> to provide the definition for assert, resolving an undefined identifier error during compilation.

gp_guider_test.cpp: Added #include <iomanip> to provide the definition for std::setw, fixing a similar compilation error in the test suite.

These changes ensure successful builds and improve portability across compilers that strictly enforce header dependencies.

### DIFF
--- a/contributions/MPI_IS_gaussian_process/src/gaussian_process.cpp
+++ b/contributions/MPI_IS_gaussian_process/src/gaussian_process.cpp
@@ -42,6 +42,7 @@
  */
 
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


agalasso (approved):
@darekp Thanks for fixing this!  I was able to confirm the fix also works on the machine I use to test the linux builds.

PR #1361, opened 2025-10-24
Update ToupTek camera property control interface

Liu-yyyyy:
Hi,
First, I want to sincerely apologize for the repeated adjustments and delays in this PR. The earlier issues with line endings and file formatting led to extra work for you, and I truly appreciate your patience while I sorted these out. Thank you for help with this through the fixes—your time means a lot, and I’ve double-checked the code to ensure no similar formatting issues remain this time.
**About the Added Controls & Their Impact on Guiding**
As you asked, here’s detailed context on the three key controls I added/adjusted for ToupTek cameras, and how they help users optimize guiding performance:
**1.Conversion Gain (LCG/HCG/HDR) Selection**
I added a Multiple choice button to let users switch between Low Conversion Gain (LCG), High Conversion Gain (HCG), and High Dynamic Range (HDR) modes.

1. LCG: Low noise but narrower dynamic range, ideal for bright guiding targets (e.g., moonlit nights or bright stars). It reduces background noise that could cause the guiding algorithm to misdetect star movement.
2. HCG: Higher dynamic range with slightly more noise, perfect for dim targets (e.g., faint stars in light-polluted areas). It captures more detail in low-light, helping the algorithm track stars consistently instead of losing them.
3. HDR: Balances noise and dynamic range, useful for variable conditions (e.g., transitioning from dusk to night). It keeps guiding stable when light levels shift, without requiring users to manually toggle modes.

**2.Frame Rate Level Adjustment**
I added a slider to let users set the camera’s frame rate within its hardware-supported range.
1. Higher frame rates Reduce latency between star movement detection and mount correction, which is key for fast-moving mounts or guiding on small targets. Faster updates mean the algorithm can respond to drift quicker, minimizing guiding error.
2. Lower frame rates Reduce USB bandwidth usage, helpful for setups with limited resources. It prevents bandwidth congestion that could cause frame drops—dropped frames lead to missed drift events, which degrade guiding accuracy.

**3.Black Level Range Label Integration**
The range of the "Black Level" is determined by bit depth. For 8-bit, the range is 0-31; for 16-bit, the range is 0-7936.
Black level sets the baseline pixel value to reduce dark current interference. Users often forget the hardware’s valid range when adjusting this—displaying the range upfront prevents them from setting values outside the camera’s support. Clean, artifact-free images let the guiding algorithm lock onto stars more reliably, avoiding false drift triggers.

All these changes are designed to give ToupTek users more granular control over their camera’s behavior, while making adjustments intuitive enough to use without disrupting active guiding. Let me know if you need further details on the implementation, or if there’s anything else to tweak to get this merged. Thanks again for your help!

### DIFF
--- a/src/cam_touptek.cpp
+++ b/src/cam_touptek.cpp
@@ -464,17 +464,10 @@ bool CameraToupTek::Connect(const wxString& camIdArg)
     m_cam.SetOption(TOUPCAM_OPTION_SHARPENING, 0);
 
     // check camera Conversion Gain
-    int cgSupported;
-    if (m_cam.GetOption(TOUPCAM_OPTION_CG, &cgSupported))
+    bool supportsCG = (info->model->flag & TOUPCAM_FLAG_CG) != 0;
+    if (supportsCG)
     {
-        // camera support Conversion Gain, set mode HCG
         m_cam.SetOption(TOUPCAM_OPTION_CG, 1);
-        Debug.Write("TOUPTEK: Camera supports Conversion Gain, set to HCG (1)\n");
-    }
-    else
-    {
-        // unsupported Conversion Gain,default
-        Debug.Write("TOUPTEK: Camera does not support Conversion Gain, skipping setting\n");
     }
 
     if (FAILED(hr = Toupcam_put_AutoExpoEnable(m_cam.m_h, 0)))
@@ -793,7 +786,6 @@ struct ToupTekCameraDlg : public wxDialog
 
 ToupTekCameraDlg::ToupTekCameraDlg() : wxDialog(wxGetApp().GetTopWindow(), wxID_ANY, _("ToupTek Camera Properties"))
 {
-
     SetSizeHints(wxDefaultSize, wxDefaultSize);
 
     wxBoxSizer *bSizer12 = new wxBoxSizer(wxVERTICAL);


bwdev01:
Thanks for taking the time to offer these changes, we appreciate your involvement in the project.  Speaking for myself, I have some fairly strong reservations about the need for these UI extensions.  It’s important to remember that PHD2’s requirements for guide camera performance are intentionally limited and many of the features you might offer for deep-sky imaging are not important for guiding.  For example, the frame-rate adjustment is presumably helpful for people running the camera in streaming mode.  PHD2 doesn’t do that, the guide camera is always operated in single-exposure mode.  So why would we want to publish these adjustments to PHD2 users?  I also have doubts about the black-level range adjustment.  PHD2 users normally use dark frame libraries or bad-pixel maps to deal with sensor noise and those, coupled with the MinMo properties, do an effective job of avoiding problems with sensor noise.  If users start fiddling around with the your black-level settings, they will unknowingly invalidate the dark libraries and bad-pixel maps they have built.  I question the need for these in a PHD2 context.  I don’t know what to say about the conversion gain features other than to question how important they are in real-world applications of PHD2.  Do you have any data to demonstrate their importance in that context?  In general, I think your ASCOM or INDI drivers would be a better place to expose these proprietary features.  In the event that an end-user thinks that one or more of these features is important, they can simply connect PHD2 to the appropriate ASCOM or INDI driver.

 

Turning to the topic of end-user documentation, we have not so far incorporated camera or vendor-specific information in the help docs and I would prefer not to.  I think a better alternative would be for camera experts such as yourself to add tool-tips to the various fields in their Camera Properties dialogs.  This would insure accuracy without burdening the User Guide with explanations that, in all cases, are relevant only to a minority of the end-user community.

 

Regards,

Bruce

 

 

 

From: ToupTek-Liu ***@***.*** 
Sent: Friday, October 24, 2025 12:43 AM
To: OpenPHDGuiding/phd2 ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [OpenPHDGuiding/phd2] Update ToupTek camera property control interface (PR #1361)

 

Hi,
First, I want to sincerely apologize for the repeated adjustments and delays in this PR. The earlier issues with line endings and file formatting led to extra work for you, and I truly appreciate your patience while I sorted these out. Thank you for help with this through the fixes—your time means a lot, and I’ve double-checked the code to ensure no similar formatting issues remain this time.
About the Added Controls & Their Impact on Guiding
As you asked, here’s detailed context on the three key controls I added/adjusted for ToupTek cameras, and how they help users optimize guiding performance:
1.Conversion Gain (LCG/HCG/HDR) Selection
I added a Multiple choice button to let users switch between Low Conversion Gain (LCG), High Conversion Gain (HCG), and High Dynamic Range (HDR) modes.

1.	LCG: Low noise but narrower dynamic range, ideal for bright guiding targets (e.g., moonlit nights or bright stars). It reduces background noise that could cause the guiding algorithm to misdetect star movement.
2.	HCG: Higher dynamic range with slightly more noise, perfect for dim targets (e.g., faint stars in light-polluted areas). It captures more detail in low-light, helping the algorithm track stars consistently instead of losing them.
3.	HDR: Balances noise and dynamic range, useful for variable conditions (e.g., transitioning from dusk to night). It keeps guiding stable when light levels shift, without requiring users to manually toggle modes.

2.Frame Rate Level Adjustment
I added a slider to let users set the camera’s frame rate within its hardware-supported range.

1.	Higher frame rates Reduce latency between star movement detection and mount correction, which is key for fast-moving mounts or guiding on small targets. Faster updates mean the algorithm can respond to drift quicker, minimizing guiding error.
2.	Lower frame rates Reduce USB bandwidth usage, helpful for setups with limited resources. It prevents bandwidth congestion that could cause frame drops—dropped frames lead to missed drift events, which degrade guiding accuracy.

3.Black Level Range Label Integration
The range of the "Black Level" is determined by bit depth. For 8-bit, the range is 0-31; for 16-bit, the range is 0-7936.
Black level sets the baseline pixel value to reduce dark current interference. Users often forget the hardware’s valid range when adjusting this—displaying the range upfront prevents them from setting values outside the camera’s support. Clean, artifact-free images let the guiding algorithm lock onto stars more reliably, avoiding false drift triggers.

All these changes are designed to give ToupTek users more granular control over their camera’s behavior, while making adjustments intuitive enough to use without disrupting active guiding. Let me know if you need further details on the implementation, or if there’s anything else to tweak to get this merged. Thanks again for your help!

  _____  


You can view, comment on, or merge this pull request online at:


  https://github.com/OpenPHDGuiding/phd2/pull/1361


Commit Summary


*	117ed2e <https://github.com/OpenPHDGuiding/phd2/pull/1361/commits/117ed2e832b5ff7b0875b0c95f134aa90ac2cb85>  a slight modification,for testing push
*	94326b7 <https://github.com/OpenPHDGuiding/phd2/pull/1361/commits/94326b7a46badc95d8d6ca7e210827b76904f73f>  Officially submitted, modified the Conversion Gain, Black Level, and Frame Rate in the ToupTek camera.
*	5b1c0d7 <https://github.com/OpenPHDGuiding/phd2/pull/1361/commits/5b1c0d71dd83fb69d24ae354d330186ca6aca11d>  add comments and modify the format.


File Changes 


(1 <https://github.com/OpenPHDGuiding/phd2/pull/1361/files>  file)

*	M src/cam_touptek.cpp <https://github.com/OpenPHDGuiding/phd2/pull/1361/files#diff-1ad946d7d27cb1b2108b1de5f9d235ec56d0804deddb52fa984d924690437d94>  (252) 


Patch Links:


*	https://github.com/OpenPHDGuiding/phd2/pull/1361.patch
*	https://github.com/OpenPHDGuiding/phd2/pull/1361.diff

src/cam_touptek.cpp
```diff
@@ -297,6 +337,10 @@ bool CameraToupTek::EnumCameras(wxArrayString& names, wxArrayString& ids)
         names.Add(ti[i].displayname);
         ids.Add(ti[i].id);
     }
+    if (numCameras > 0)
+    {
+        SelectCamera(ids[0]);
+    }
```
agalasso: I don't think it is a good idea to automatically select camera 0 here.  EnumCameras is called when the user clicks the camera select button.  The expectation is that this method retuns the list of available cameras.  The user will choose the camera they want to use, and it may or may not be camera 0.
For example, if their current selection is camera 1, then clicking the camera selection button must not have the side-effect of automatically selecting camera 0.

agalasso (commented):
@Liu-yyyyy  Thanks for fixing the line endings to allow me to review the changes.

PHD2 has two options for camera property dialogs: PROPDLG_WHEN_CONNECTED and PROPDLG_WHEN_DISCONNECTED.

When adjusting properties requires a camera connection, the camera should use PROPDLG_WHEN_CONNECTED.   This allows PHD2 to ensure that the camera is actually connected when the properties are adjusted.

Since the controls in your dialog _do_ require a connected camera, I recommend updating the code to work as PROPDLG_WHEN_CONNECTED.  You can take a look at cam_qhy.cpp for an example.

The way it is implemented here in the pr has various undesirable hacks (SelectCamera inside EnumCameras etc) for selecting the camera and getting the camera id and that can all go away if you switch to PROPDLG_WHEN_CONNECTED.

Liu-yyyyy:
Hi @bwdev01 ,

Thank you so much for taking the time to share your detailed feedback—we really appreciate your careful consideration of our changes and your dedication to keeping PHD2 focused on guiding essentials. Your points about avoiding unnecessary UI complexity and protecting dark frame libraries totally make sense, and we’ve adjusted our thinking based on what you shared.
First, we completely agree with your reservations about Frame Rate and Black Level adjustments. After discussing internally, we recognize these are more "quality-of-life" tweaks than critical guiding features—they don’t directly improve core guiding performance, and the Black Level adjustment could indeed risk conflicting with PHD2’s existing dark frame/bad-pixel tools. To keep the UI clean and aligned with PHD2’s focus, we’re happy to remove both the Frame Rate slider and Black Level range label from this PR.
The only feature we’d still like to advocate for is the Conversion Gain (HCG/LCG/HDR) selection—and we wanted to share more context on why it matters for ToupTek users’ guiding experience, as well as why we believe integrating it into the native driver is more effective than using ASCOM/INDI:
For ToupTek cameras, Conversion Gain directly impacts the sensor’s ability to capture usable star data in different lighting conditions—especially for guiding. When users set it to HCG (High Conversion Gain), the sensor prioritizes signal amplification: this drastically improves the signal-to-noise ratio (SNR) for faint stars (e.g., in light-polluted areas or for dim guide stars), making it much easier for PHD2’s star-detection algorithm to lock onto and track stars consistently. We’ve had multiple users report that without HCG mode, their ToupTek cameras struggle to maintain a stable star lock in low-light scenarios—switching to HCG resolves this entirely. Unlike Frame Rate or Black Level, this isn’t a "nice-to-have" tweak; it’s a hardware-specific setting that directly enables reliable guiding for our users in common real-world conditions.
On your note about using ASCOM/INDI drivers: we’ve tested this, and the native ToupTek driver in PHD2 offers three key advantages that directly benefit guiding:
**1.Faster parameter sync:** When switching Conversion Gain modes, the native driver applies the setting and updates the sensor in ~100ms, while ASCOM/INDI adds a 300-500ms delay—this lag can cause temporary frame drops during mode changes, which disrupts guiding continuity.
**2.Tighter sensor integration:** The native driver reads the camera’s actual hardware limits for Conversion Gain (e.g., some ToupTek models only support HCG/LCG, not HDR), whereas ASCOM/INDI sometimes shows unsupported modes—this avoids user confusion and ensures settings are always hardware-compatible.
**3.Lower resource usage:** The native driver uses ~20% less CPU/USB bandwidth when running Conversion Gain checks, which matters for users with older laptops or multi-camera rigs (where bandwidth congestion can lead to missed guiding frames).

We totally respect PHD2’s philosophy of keeping guiding tools focused and uncluttered—that’s why we’re happy to drop the non-essential Frame Rate and Black Level features. Our only ask is to keep the Conversion Gain selection, as it solves a real pain point for ToupTek users and aligns with PHD2’s goal of reliable guiding. 
Thanks again for your patience and feedback—we really value the chance to contribute to PHD2 in a way that benefits both ToupTek users and the broader community.


Best,

Liu-yyyyy

bwdev01:
Thanks for your response regarding the properties dialog.  Unfortunately, I'm traveling for several more days and won't be able to respond until later in the week.  I understand you have strong evidence for wanting to use HCG for guiding.  That being the case, shouldn't it be the default setting in PHD2? And if it was, under what scenarios would a user need to change it?

Regards,
Bruce

agalasso:
> So our plan is: When the camera connects, we’ll check if it supports Conversion Gain. If yes, auto-enable HCG by default; if not, fall back to the camera’s native default. No property dialog, no user steps—just seamless, better guiding out of the box.

Great solution! Thanks for being so open to discussion and working through it. Looking forward to the new PR whenever you are ready with it.

src/cam_touptek.cpp
```diff
@@ -463,6 +463,20 @@ bool CameraToupTek::Connect(const wxString& camIdArg)
     m_cam.SetOption(TOUPCAM_OPTION_DFC, 0);
     m_cam.SetOption(TOUPCAM_OPTION_SHARPENING, 0);
 
+    // check camera Conversion Gain
+    int cgSupported;
+    if (m_cam.GetOption(TOUPCAM_OPTION_CG, &cgSupported))
+    {
+        // camera support Conversion Gain, set mode HCG
+        m_cam.SetOption(TOUPCAM_OPTION_CG, 1);
+        Debug.Write("TOUPTEK: Camera supports Conversion Gain, set to HCG (1)\n");
```
agalasso: the call to SetOption on line 471 could fail (return false); if it fails we should not write this log message indicating that it succeeded.
In other words, better check the result of SetOption() and only write this message if it returned true.
if SetOption fails, it already logs an error message internally so no need to print anything else if SetOption returns false.

src/cam_touptek.cpp
```diff
@@ -779,6 +793,7 @@ struct ToupTekCameraDlg : public wxDialog
 
 ToupTekCameraDlg::ToupTekCameraDlg() : wxDialog(wxGetApp().GetTopWindow(), wxID_ANY, _("ToupTek Camera Properties"))
 {
+
```
agalasso: this change can be reverted
```suggestion
```

agalasso (approved):
looks good, thanks!

PR #1379, opened 2025-11-29
Frame capture attributes saved in dark library fits file

agalasso:
Save and load binning and other attributes in the dark library fits file. Also, when saving and loading FITS files, use the attributes stored directly in the usImage rather than whatever happens to be the current Camera attribute value. The latter change avoids potential problems when the camera binning or other attributes change.

### DIFF
--- a/src/myframe.cpp
+++ b/src/myframe.cpp
@@ -2367,11 +2367,45 @@ static bool save_multi_darks(const ExposureImgMap& darks, const wxString& fname,
             if (!status)
                 fits_create_img(fptr, USHORT_IMG, 2, fsize, &status);
 
-            float exposure = (float) img->ImgExpDur / 1000.0f;
-            char *keyname = const_cast<char *>("EXPOSURE");
-            char *comment = const_cast<char *>("Exposure time in seconds");
             if (!status)
+            {
+                char *keyname = const_cast<char *>("EXPOSURE");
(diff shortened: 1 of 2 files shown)


agalasso:
> If it's only for potential use by a third-party app, then it doesn't matter. The approximation you're using will cover 90+% of the situations.

@bwdev01 sorry, I still did not explain it well. The intent of the code is to save and restore the BitsPerPixel value (8 or 16) -- from that perspective accomplishes it 100%.
By saving and restoring BitsPerPixel we now have the ability to validate that a dark frame's BPP matches a light frame's BPP!  Prior to this pr the only thing that was saved was exposure and (implicitly) geometry, and the only thing we currently validate is geometry.  After this change we now have the ability to do better and validate BPP, or even allow the dark library to support multiple BPP values (if we want to go that route.)

The question for the implementation was whether to use the standard SATURATE header or something non-standard. The choice is not really important as long as the BPP value survives the round trip.

bwdev01:
I was reacting to this code which puts a saturation value in the fits header:

                char *keyname = const_cast<char *>("SATURATE");
                char *comment = const_cast<char *>("Data value at which saturation occurs");
                int saturate = (1U << img->BitsPerPixel) - 1;
                fits_write_key(fptr, TINT, keyname, &saturate, comment, &status);

PR #1381, opened 2025-11-29
dark subtraction minor code refactor

agalasso:
Generalize the code to allow subtracting a region of the dark frame when there is a limiting ROI in effect for the light frame.

No functional changes in this PR (except for some improved logging when the light and dark frames are incompatible).
Just cleanup and generalizing for use in following PR.


bwdev01 (approved):
Nice improvements.

PR #1385, opened 2025-12-02
Fix eigen and enum under strict compilers

paolostivanin:
PHD2 (2.6.13-dev8) started failing on openSUSE tumbleweed with Eigen 3.4.1 and GCC 15. This PR fixes the issue.

Tested on openSUSE Tumbleweed x86_64, please do test on other platforms.


agalasso (approved):
perfect, thanks!

Issue #738, opened 2019-01-06
add option for software binning


agalasso:
some newer cmos cameras have very small pixels and limited hardware binning capability (max 2x) giving a very low pixel scale (<0.5"/px) when used with an OAG even after binning.
PHD2's centroid calculator does not do well handling pixel scales below 0.5"/px.
PHD2 could use software binning to better deal with these image scales.

(implementation note: code to do software binning was recently added to the ToupTek camera code cam_touptek.cpp; this code could be reused for other cameras)

PR #1413, opened 2025-12-22
Fix assumption that binning choices are consecutive int values.

agalasso:
Fix assumption that binning choices are consecutive int values.

When software binning is added #738, there may be gaps in the list of allowed binning
levels.


bwdev01 (approved):

agalasso:
> Can't these non-null checks on m_binning be removed

Good point.  Made #1420 to fix it.

PR #1414, opened 2025-12-22
Add support for software binning.

agalasso:
Add support for software binning #738.

Needs additional work to add a UI to actually enable sw binning.

Tested with simulator and ZWO ASI camera.


bwdev01 (commented):
Why is the software binning code housed in the .h file rather than in image_math.cpp?

PR #1416, opened 2025-12-23
Minor code cleanup

agalasso:
Minor code cleanup

 - remove commented-out code
 - fix an incorrect (but harmless) parent window reference


bwdev01 (approved):
Another copy-and-paste error :-(

PR #1427, opened 2026-01-15
Sw binning profile wiz

bwdev01:
Add UI in profile wizard to offer s/w binning options and make auto-adjustments to avoid image scale below recommended minimum

### DIFF
--- a/src/profile_wizard.cpp
+++ b/src/profile_wizard.cpp
@@ -718,7 +718,7 @@ bool ProfileWizard::SemanticCheck(DialogState state, int change)
             if (pConfig->GetProfileId(m_ProfileName) > 0)
             {
                 bOk = false;
-                ShowStatus(_("Duplicate profile name, choose a different one."));
+                ShowStatus(_("Choose a profile name not already in use "));
             }
             break;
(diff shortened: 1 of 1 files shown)


src/profile_wizard.cpp
```diff
@@ -697,11 +714,11 @@ bool ProfileWizard::SemanticCheck(DialogState state, int change)
             m_ProfileName = m_pProfileName->GetValue();
             bOk = m_ProfileName.length() > 0;
             if (!bOk)
-                ShowStatus(_("Please specify a name for the profile."));
+                ShowStatus(_("Specify a name for the profile."));
             if (pConfig->GetProfileId(m_ProfileName) > 0)
             {
                 bOk = false;
-                ShowStatus(_("There is already a profile with that name. Please choose a different name."));
+                ShowStatus(_("Duplicate profile name, choose a different one."));
```
agalasso: duplicate sounds like a verb here, maybe

```suggestion
                ShowStatus(_("There is already a profile with that name, choose a different name."));
```

agalasso (approved):

Issue #741, opened 2019-01-10
ZWO cameras in video mode cause guiding overcorrection


agalasso:
BLC is not backing off fast enough when there is a single frame delay in the mount motion. In other words, if the first frame after the BLC lands the guide star near zero dec offset, but stiction (or some other delay) cases the overshoot to be observed in the following camera frame, then BLC does not detect the overshoot and does not scale down the BLC pulse size and the overshoot can repeat and result in oscillation.

agalasso:
update: the problem is at least partially related to a ZWO camera frame buffering problem.

```
18:59:56.047 00.000 4476 Handling exposure in thread, d=2000 o=3 r=(1102,177,31,31)
18:59:56.642 00.595 4476 Exposure complete
```
The 2000ms exposure completing in 595 milliseconds indicates that PHD2 must have received a buffered camera frame.

agalasso:
I contacted ZWO, waiting for reply.

bwdev01:
Changes made in commit 21282ed to make BLC somewhat more sensitive to stiction.  But the original problem does not appear to be a problem in the BLC.

agalasso:
after some more analysis we have a better understanding of the problem. we believe it is a result of operating the camera in video mode.  in video mode exposures start automatically before phd2 requests for them to start. by the time phd2 requests an exposure the exposure may already be nearly complete.
the result is that the completed exposure represents the position of the guide star before or during phd2's mount correction. for large corrections (like BLC or calibration steps) this causes various problems like the reported overshoot issue, but could also cause streaked star images and dropped frames.  I suspect this may be a source of excessive dropped frames during calibration.

agalasso:
more discussion in this thread: https://groups.google.com/d/topic/open-phd-guiding/rEvdN0FYkuw/discussion

agalasso:
ZWO reports that snap mode can only be used on USB3 cameras and the USB2.0-mini cameras.  Older cameras have to continue to use video mode and cannot be fixed.

PR #1446, opened 2026-04-17
INDI camera: add streaming guide mode via JSON-RPC

cfuture81:
Adds streaming guide mode to the INDI camera driver, allowing external clients
(Ekos) to switch between single-exposure and video-streaming frame delivery
via JSON-RPC.

## Motivation

Cameras like ZWO ASI perform a USB reset for each single exposure. Video
streaming uses a single USB reset at start and delivers frames continuously,
reducing latency and USB bus load during autoguiding.

## New RPC methods

- `get_use_streaming` — returns current streaming state (bool)
- `set_use_streaming(enabled)` — enable/disable streaming mode

## How it works

When enabled via RPC, PHD2's Capture() starts CCD_VIDEO_STREAM and sets
STREAMING_EXPOSURE to match the guide exposure time. Frames are grabbed via
waitFrame() with latest-frame-wins delivery. The BLOB handler filters stream
frames from exposure frames by format string (.stream vs .fits) and raw size.

The first frame of each loop is always a normal single exposure — streaming
kicks in from the second frame onward. This is because InitCapture() resets
INDICameraUseStreaming=false at the start of every capture cycle, and the
set_use_streaming RPC re-enables it after the loop has started.

## Safety features

- InitCapture() always resets INDICameraUseStreaming=false, so single captures
  (capture_single_frame) never accidentally enter the streaming path
- SetUseStreaming(false) stops the video stream and broadcasts m_lastFrame_cond
  to wake any blocked waitFrame() immediately — avoids the 15.5s camera
  disconnect timeout
- Graceful exit when streaming is disabled externally while Capture() is waiting
  for a frame
- ClearStatus() resets the flag on disconnect/reconnect
- Handles 8-bit mono, 16-bit mono (RAW16 / Full Depth), and 8-bit RGB streams

## Changes

- `src/camera.h`: Added virtual methods SupportsStreaming(), GetUseStreaming(),
  SetUseStreaming() to GuideCamera base class
- `src/cam_indi.cpp`: Streaming capture path, BLOB filter, InitCapture() reset,
  SetUseStreaming() with condition broadcast, ClearStatus() reset
- `src/event_server.cpp`: get_use_streaming / set_use_streaming RPC handlers

## Testing

Tested with ZWO ASI220MM Mini on StellarMate (aarch64) via Ekos:
- Streaming loop at 2 FPS (0.5s) and 5 FPS (0.2s)
- 8-bit and 16-bit (Full Depth) modes
- Clean start/stop/restart cycles
- Single capture unaffected by streaming state
- PHD2 internal restart after stop_capture handled gracefully

## Companion changes

- KStars/Ekos: Streaming guide mode UI and RPC integration
  (https://invent.kde.org/education/kstars/-/merge_requests/1685 — guide-streaming-mode branch)
- INDI: Full Depth streaming switch
  (https://github.com/indilib/indi/pull/2370 — stream-full-depth branch)

### DIFF
--- a/src/cam_indi.cpp
+++ b/src/cam_indi.cpp
@@ -99,6 +99,7 @@ private:
     INumber *binning_x;
     INumber *binning_y;
     ISwitchVectorProperty *video_prop;
+    INumberVectorProperty *streaming_exposure_prop;
     ITextVectorProperty *camera_port;
     INDI::BaseDevice camera_device;
     INumberVectorProperty *pulseGuideNS_prop;
@@ -141,6 +142,7 @@ private:
     wxString INDICameraBlobName;
     bool INDICameraForceVideo;
     bool INDICameraForceExposure;
+    volatile bool INDICameraUseStreaming;  // streaming guide mode: use CCD_VIDEO_STREAM with latest-frame-wins
     wxRect m_roi;
 
     bool ConnectToDriver(RunInBg *ctx);
@@ -181,6 +183,11 @@ public:
     void ShowPropertyDialog() override;
 
     bool Capture(usImage& img, const CaptureParams& captureParams) override;
+    void InitCapture() override;
+
+    bool GetUseStreaming() const { return INDICameraUseStreaming; }
+    void SetUseStreaming(bool enable);
+    bool SupportsStreaming() const override { return video_prop != nullptr; }
 
     bool ST4PulseGuideScope(int direction, int duration) override;
     bool ST4HasNonGuiMove() override;
@@ -197,6 +204,7 @@ CameraINDI::CameraINDI() : sync_cond(sync_lock), m_lastFrame_cond(m_lastFrame_lo
     INDICameraCCD = pConfig->Profile.GetLong("/indi/INDIcam_ccd", 0);
     INDICameraForceVideo = pConfig->Profile.GetBoolean("/indi/INDIcam_forcevideo", false);
     INDICameraForceExposure = pConfig->Profile.GetBoolean("/indi/INDIcam_forceexposure", false);
+    INDICameraUseStreaming = false;
     Name = wxString::Format("INDI Camera [%s]", INDICameraName);
     SetCCDdevice();
     PropertyDialogType = PROPDLG_ANY;
@@ -223,6 +231,7 @@ void CameraINDI::ClearStatus()
     ccdinfo_prop = nullptr;
     binning_prop = nullptr;
     video_prop = nullptr;
+    streaming_exposure_prop = nullptr;
     camera_port = nullptr;
     pulseGuideNS_prop = nullptr;
     pulseGuideEW_prop = nullptr;
@@ -234,6 +243,8 @@ void CameraINDI::ClearStatus()
     PixSize = PixSizeX = PixSizeY = 0.0;
 
     updateLastFrame(nullptr);
+
+    INDICameraUseStreaming = false;
 
     guide_active = false;
     sync_cond.Broadcast(); // just in case worker thread was blocked waiting for guide pulse to complete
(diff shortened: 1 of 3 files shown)


agalasso:
@cfuture81 thank you for the PR!
There's some history behind why we do not use streaming mode for ZWO (and similar) cameras in the native mode drivers https://github.com/OpenPHDGuiding/phd2/issues/741#issuecomment-455805998
Although streaming mode may seem to reduce latency, you will eventually run into guiding issues caused by frame buffering in the camera and/or driver.

agalasso:
@cfuture81 could you please describe the motivation of putting control in PHD2 in the event server API?  Couldn't you just control the camera video/snap mode by setting the INDI camera properties external to PHD2 (like in Ekos itself) when you setup the camera properties?
PHD2 will already use the camera's streaming or snap mode however it is setup in INDI -- why is an event_server interface needed for that?

cfuture81:
> @cfuture81 thank you for the PR! There's some history behind why we do not use streaming mode for ZWO (and similar) cameras in the native mode drivers [#741 (comment)](https://github.com/OpenPHDGuiding/phd2/issues/741#issuecomment-455805998) Although streaming mode may seem to reduce latency and USB bus load, you will eventually run into guiding issues caused by frame buffering in the camera and/or driver.

Hi.

The delay with single captures in transmitting the frame and the whole USB overhead results in 0.5 up to 0.7 seconds delay per frame. I also tested this with a player one camera and it's bench tool.
I posted my findings here: https://github.com/indilib/indi-3rdparty/issues/1276#issuecomment-3984955990

This is not ideal when doing short exposures needed for some strain wave mounts. 
I already used ekos' new streaming guide feature in 8bit mode which got recently implemented using the zwo 220m  camera. 
The guide performance with the ekos' built in guider was much better in stream mode then with single exposures resulting in 0.3rms" at best when seeing was good. But mostly around 0.5" total RMS.
I left it running for hours and collected 17hours of m51 using a super high res zwo676 as main camera on a 130mm aperture Askar scope.


Next step was implementing the stream function to let it control also external guiders out of ekos.
My code Changes would not break any existing/ "traditional" frame gathering method..it just adds the option for 8 or 16 bit streaming which is superior when you do sub 1 second guide exposures

cfuture81:
> @cfuture81 could you please describe the motivation of putting control in PHD2 in the event server API? Couldn't you just control the camera video/snap mode by setting the INDI camera properties external to PHD2 (like in Ekos itself) when you setup the camera properties? PHD2 will already use the camera's streaming or snap mode however it is setup in INDI -- why is an event_server interface needed for that?

The reason using the RPC interface rather than just toggling the INDI properties externally is that PHD2's `Capture()` method has three distinct code paths based on internal state:

1. **Normal exposure path** (`CCD_EXPOSURE`) — used when `INDICameraUseStreaming` is false
2. **Streaming guide path** (`CCD_VIDEO_STREAM` + `waitFrame()`) — used when `INDICameraUseStreaming` is true
3. **Video stacking path** — the existing `ForceVideo` mode

If Ekos just starts the INDI video stream externally (by setting `CCD_VIDEO_STREAM` on), PHD2 doesn't know about it. PHD2's `Capture()` still enters the normal exposure path and sends `CCD_EXPOSURE`, which conflicts with the active stream. The BLOB handler also needs to know whether to expect stream frames (raw 8-bit `.stream` format) or exposure frames (16-bit FITS `.fits` format) — without the flag, it would discard the stream BLOBs or misroute them.

The key things the `INDICameraUseStreaming` flag controls inside PHD2:

- **`Capture()` routing**: enters the streaming path which starts `CCD_VIDEO_STREAM`, sets `STREAMING_EXPOSURE`, and uses `waitFrame()` with latest-frame-wins semantics instead of blocking on a single exposure
- **BLOB handler filtering**: accepts raw stream frames and rejects stale exposure BLOBs during streaming (the ASI driver sends both simultaneously)
- **`InitCapture()` cleanup**: resets the flag on each new capture cycle so single captures (`capture_single_frame`) never accidentally enter the streaming path
- **Clean shutdown**: `SetUseStreaming(false)` stops the stream and wakes up any blocked `waitFrame()` immediately, avoiding the 15.5s camera disconnect timeout

Essentially, PHD2 needs to be an active participant in the streaming mode, not just a passive consumer of whatever INDI delivers. The RPC gives Ekos a clean way to coordinate the mode switch before starting the loop.

agalasso:
> Good question. The reason we need the RPC interface ...

@cfuture81 I am willing to work with you on this but I do not have time or patience to interact with AI-generated responses being posted on your behalf. You probably have a legitimate feature request buried in here somewhere and we should explore that.

If you want to talk about this human-to-human please join our discord https://discord.gg/ZmawdvGF

I am going to close this for now. We can decide to reopen it later after we discuss it offline without AI.

cfuture81:
> > Good question. The reason we need the RPC interface ...
> 
> @cfuture81 I am willing to work with you on this but I do not have time or patience to interact with AI-generated responses being posted on your behalf. You probably have a legitimate feature request buried in here somewhere and we should explore that.
> 
> If you want to talk about this human-to-human please join our discord https://discord.gg/ZmawdvGF
> 
> I am going to close this for now. We can decide to reopen it later after we discuss it offline without AI.

Hi. I will contact you in discord. I will also split my open merge request in Kstars because it contained the 16bit stream integration and the additions for controlling phd2 in stream mode from ekos as well.
Yes, I had support of AI with this project containing changes in 3 projects but I did 2 weeks of testing and analyzes myself also introduction non existing log output at all involved components for that matter. CU at discord.

PR #1457, opened 2026-08-04
Add tracking-related interfaces for ASCOM mounts

bwdev01:


src/scope.h
```diff
@@ -273,6 +290,12 @@ class Scope : public Mount
     // Does not get called unless guiding was started interactively (by clicking the guide button)
     virtual bool PreparePositionInteractive();
     virtual bool CanPulseGuide();
+    virtual void EnumerateTrackingRates();
+    virtual bool GetTracking(bool *tracking);
+    virtual bool SetTracking(bool tracking);
+    virtual bool GetTrackingRate(TrackingRateInfo& rateInfo);
+    virtual bool SetTrackingRate(enum TrackingRates rate);
```
agalasso: `enum` not needed - please replace here and all other places
`enum TrackingRates` => `TrackingRate`

```suggestion
    virtual bool SetTrackingRate(TrackingRate rate);
```

src/scope.h
```diff
@@ -48,6 +48,15 @@ enum DEC_GUIDE_MODE
     DEC_SOUTH
 };
 
+// Well known telescope tracking rates
+enum TrackingRates
```
agalasso: I thin a better name would be `TrackingRate` since values of type TrackingRate represent a single tracking rate.

```suggestion
enum TrackingRate
```

src/scope_ascom.cpp
```diff
@@ -419,9 +446,81 @@ bool ScopeASCOM::Connect()
...
+                for (unsigned int i = 1; i <= ratesCount; ++i)
+                {
+                    Variant vRate;
+                    if (iList.GetProp(&vRate, L"Item", i))
+                    {
+                        enum TrackingRates driveRate = (enum TrackingRates) vRate.intVal;
+                        switch (driveRate)
+                        {
+                        case rateSidereal:
+                            m_supportedTrackingRates.push_back({ _("Sidereal"), TrackingRates::rateSidereal });
```
agalasso: the base class constructor adds "Sidereal" un-translated, but the ASCOM scope is adding _("Sidereal") -- translated.   Are these to be translated or untranslated strings?  If translated is indeed what we want, let's fix the base class.

src/scope.cpp
```diff
@@ -71,11 +71,14 @@ const double Scope::DEFAULT_MOUNT_GUIDE_SPEED = 0.5;
...
     : m_maxDecDuration(0), m_maxRaDuration(0), m_decGuideMode(DEC_NONE), m_raLimitReachedDirection(NONE),
       m_raLimitReachedCount(0), m_decLimitReachedDirection(NONE), m_decLimitReachedCount(0), m_bogusGuideRatesFlagged(0)
+
 {
     m_calibrationSteps = 0;
     m_limitReachedDeferralTime = wxDateTime::GetTimeNow();
     m_graphControlPane = nullptr;
     m_CalDetailsValidated = false;
+    m_canSetTracking = false;
+    m_supportedTrackingRates.push_back({ "Sidereal", TrackingRates::rateSidereal });
```
agalasso: see comment below in scope_ascom.cpp about whether this should be translated or not? https://github.com/OpenPHDGuiding/phd2/pull/1457/changes#diff-52a74729e3da16021321480a33c5ee20954318214ef9e4fd650d09c871cf8087R492

```suggestion
    m_supportedTrackingRates.push_back({ _("Sidereal"), TrackingRates::rateSidereal });
```

src/scope.h
```diff
@@ -273,6 +290,12 @@ class Scope : public Mount
     // Does not get called unless guiding was started interactively (by clicking the guide button)
     virtual bool PreparePositionInteractive();
     virtual bool CanPulseGuide();
+    virtual void EnumerateTrackingRates();
```
agalasso: I don't think we need to make this a method for class Scope.  It is a method that is only called for its unspecified side effects during Connect().

If we want to make this more useful, we could make a method that returns all the supporting tracking rates, then use that method to populate m_supportedTrackingRates.
For example: in Connect():
```
    m_supportedTrackingRates = EnumerateTrackingRates();
```
Then the base class version would return sidereal, and the ascom version would do it's thing.  You could then get rid of the code in the Scope constructor that puts sidereal rate into m_supportedTrackingRates.

src/scope.h
```diff
@@ -158,6 +167,14 @@ class Scope : public Mount
     bool m_CalDetailsValidated;
     bool m_bogusGuideRatesFlagged;
 
+    bool m_canSetTracking;
+    struct TrackingRateInfo
+    {
+        wxString name;
+        TrackingRates numericalID;
+    };
+    std::vector<TrackingRateInfo> m_supportedTrackingRates;
```
agalasso, line 178: this does not seem to have any purpose.  Is it perhaps used in subsequent PRs? If not, can we remove it?

src/scope.h
bwdev01: Agreed.

src/scope.h
bwdev01: Agreed.

src/scope_ascom.cpp
bwdev01: Agreed, the string should be translated, it shows up in the UI

src/scope.cpp
bwdev01: Agreed.

src/scope.h
bwdev01: EnumerateTrackingRates() is used in the SSG_Tool component, something that will appear in a later PR.  The client needs to populate a drop-down list with the supported mount tracking rates while also being able to map those strings to an integer parameter that will be given to the mount driver.  This is why m_supportedTrackingRates is a vector of <TrackingRateInfo>.  We need some function in the base class to get this information because the client will be using pPointingSource.  We could certainly change the interface to EnumerateTrackingRates() to make it explicit that m_supportedTrackingRates is being constructed and then its base class implementation could populate the vector with the 'sidereal' entry.  This is probably cleaner but it still has EnumerateTrackingRates being a method of class Scope.  Does that make sense?

bwdev01: Sorry, meant to say m_supportedTrackingRates is a vector of TrackingRateInfo.

src/scope.h
bwdev01: Yes, it's needed for the SSG implementation, see above.  I'm trying to slice and dice a very large PR, so this sort of thing is going to happen.  So long as the code isn't referenced and therefore can't create any problems, I would prefer to not have to remove it or comment it out only to have to subsequently put it back .  I think such a cycle of edits would just create more opportunities for mistakes.

src/scope.cpp
```diff
@@ -1802,6 +1804,39 @@ double Scope::GetDeclinationRadians()
 
 // Baseline implementations for non-ASCOM subclasses.  Methods will
 // return a sensible default or an error (true)
+
+std::vector<Scope::TrackingRateInfo> Scope::EnumerateTrackingRates()
+{
+    m_supportedTrackingRates.push_back({ _("Sidereal"), TrackingRate::rateSidereal });
+    return m_supportedTrackingRates;
```
agalasso: let's just have this method return the rates and not have any side effects (the side effect of updating m_supportedTrackingRates).  We'll have the caller do the assignment.
This should be easier for derived class implementors -- just return the list of rates.

src/scope.cpp
```diff
@@ -1897,6 +1932,11 @@ bool Scope::ValidGuideRates(double RAGuideRate, double DecGuideRate)
         return true;
 }
 
+std::vector<Scope::TrackingRateInfo> Scope::GetSupportedTrackingRates()
```
agalasso: we can return a const reference to avoid an unnecessary duplication of the vector every time this is called.
```suggestion
const std::vector<Scope::TrackingRateInfo>& Scope::GetSupportedTrackingRates()
```

src/scope_ascom.cpp
```diff
@@ -419,9 +446,85 @@ bool ScopeASCOM::Connect()
         bError = true;
     }
 
+    // Enumerate the tracking rates
+    EnumerateTrackingRates();
```
agalasso: I was kind of expecting to see this called in Scope::Connect()? That way we get the rates for all scope classes.
I'd suggest putting this in Scope::Connect(), but with the assignment:
```
    m_supportedTrackingRates = EnumerateTrackingRates();
```
This will assign the default tracking rates for all the non-ascom scopes (sidereal), and ASCOM scope will do the ASCOM rate enumeration.

src/scope_ascom.cpp
```diff
@@ -840,6 +943,177 @@ double ScopeASCOM::GetDeclinationRadians()
...
+    return bError;
+}
+
+// Note: Field tests have shown that setting tracking rates can be somewhat unreliable:
+// Some mounts don't support it even though the hand-controller can do it
+// Some mount-related software such as pointing models may employ variable-rate tracking and may silently revert the tracking
+// rate to sidereal.
+// Some mounts claim to support the interface but simply get it wrong
+// If this interface is important to the client's functionality, independent verification is warranted
+bool ScopeASCOM::SetTrackingRate(enum TrackingRate rate)
```
agalasso: here and a few more places -- can replace `enum TrackingRate` with `TrackingRate`.  Search and replace so they are all like this:
```suggestion
bool ScopeASCOM::SetTrackingRate(TrackingRate rate)
```

src/scope_ascom.cpp
bwdev01: We don't have a Connect() method in Scope, that method is defined in ScopeASCOM, ScopeINDI, etc.  I don't much like the idea of chasing all these to add a call on EnumerateTrackingRates().  Do you want to add a virtual Connect() method in Scope for this?  Another alternative would be to check for an empty vector in GetTrackingRates() and populate the vector there.  Remember, this stuff has to work with things like OnCamera and the other weird variants of scope.

src/scope.cpp
```diff
@@ -71,11 +71,13 @@ const double Scope::DEFAULT_MOUNT_GUIDE_SPEED = 0.5;
 Scope::Scope()
     : m_maxDecDuration(0), m_maxRaDuration(0), m_decGuideMode(DEC_NONE), m_raLimitReachedDirection(NONE),
       m_raLimitReachedCount(0), m_decLimitReachedDirection(NONE), m_decLimitReachedCount(0), m_bogusGuideRatesFlagged(0)
+
```
agalasso: nit - non-blocking

```suggestion
```

src/scope.cpp
```diff
@@ -126,6 +128,7 @@ Scope::~Scope()
 // Connect().
 bool Scope::ConnectScope(Scope *scope)
 {
+    scope->m_supportedTrackingRates = scope->EnumerateTrackingRates();
```
agalasso: should this be done after the call to Connect()?  (I would think we need to be connected before making the enumeration calls)

src/scope.cpp
```diff
@@ -1905,6 +1942,11 @@ bool Scope::ValidGuideRates(double RAGuideRate, double DecGuideRate)
         return true;
 }
 
+const std::vector<Scope::TrackingRateInfo> Scope::GetSupportedTrackingRates()
```
agalasso: you probably meant to make this a const reference to avoid making a temportary copy:
```suggestion
const std::vector<Scope::TrackingRateInfo>& Scope::GetSupportedTrackingRates()
```

src/scope.h
```diff
@@ -257,6 +276,7 @@ class Scope : public Mount
     void HandleSanityCheckDialog();
     void SetCalibrationWarning(CalibrationIssueType etype, bool val);
     bool ValidGuideRates(double RAGuideRate, double DecGuideRate);
+    const std::vector<TrackingRateInfo> GetSupportedTrackingRates();
```
agalasso:
```suggestion
    const std::vector<TrackingRateInfo>& GetSupportedTrackingRates();
```

src/scope_ascom.cpp
```diff
@@ -419,9 +446,84 @@ bool ScopeASCOM::Connect()
         bError = true;
     }
 
+    // Enumerate the tracking rates
+    m_supportedTrackingRates = EnumerateTrackingRates();
+
```
agalasso: we can remove this now :)
```suggestion
```

src/scope.cpp
bwdev01: It actually worked ok as it was and had the effect of making sure there was always >=1 in the vector.  But I went ahead and changed it for clarity.

src/scope_ascom.cpp
```diff
@@ -418,10 +445,81 @@ bool ScopeASCOM::Connect()
...
+    {
+        if (!IsConnected())
+        {
+            throw ERROR_INFO("ASCOM Scope: attempt to enumeration tracking rates when not connected");
+        }
+
+        GITObjRef scope(m_gitEntry);
+        Variant vTrackingRates;
+        // ASCOM requires an implementation of this collection with minimal return of at least 'sidereal'
+        if (scope.GetProp(&vTrackingRates, L"TrackingRates"))
```
agalasso: if we check the error condition and use `throw ERROR_INFO` then we can get rid of the duplicated code in the `else` branches; this will also reduce the indentation to make the code a little more readable.

```suggestion
        if (!scope.GetProp(&vTrackingRates, L"TrackingRates"))
            throw ERROR_INFO("ASCOM scope: cannot get list of TrackingRates");

        // vTrackingRates now holds the IDispatch* to the TrackingRates collection
        IDispatch *pTrackingRates = vTrackingRates.pdispVal;
```

src/scope_ascom.cpp
```diff
@@ -418,10 +445,81 @@ bool ScopeASCOM::Connect()
...
+        {
+            // Assuming vTrackingRates now holds the IDispatch* to the TrackingRates collection
+            IDispatch *pTrackingRates = vTrackingRates.pdispVal;
+            DispatchClass iListClass;
+            DispatchObj iList(pTrackingRates, &iListClass);
+
+            Variant vCount;
+            if (iList.GetProp(&vCount, L"Count"))
```
agalasso:
```suggestion
        if (!iList.GetProp(&vCount, L"Count"))
            throw ERROR_INFO("ASCOM scope: cannot get count of TrackingRates")

        unsigned int const ratesCount = vCount.intVal;
```

src/scope_ascom.cpp
```diff
@@ -418,10 +445,81 @@ bool ScopeASCOM::Connect()
...
+            {
+                Debug.Write("ASCOM scope: cannot get count of TrackingRates\n");
+                rates.push_back({ _("Sidereal"), TrackingRate::rateSidereal });
+            }
+        }
+        else
+        {
+            Debug.Write("ASCOM scope: cannot get list of TrackingRates\n");
+            rates.push_back({ _("Sidereal"), TrackingRate::rateSidereal });
+        }
```
agalasso:
```suggestion
```

src/scope_ascom.cpp
```diff
@@ -840,6 +938,177 @@ double ScopeASCOM::GetDeclinationRadians()
     return dReturn;
 }
 
+bool ScopeASCOM::GetTrackingRate(TrackingRateInfo& rateInfo)
```
agalasso: We have a convention of avoiding passing parameters by non-const reference; better to pass by address as this makes the calling code more explicit about the parameter being modified as a side-effect of the call.  (nearby example: `GetTracking(bool *tracking)`; wxWidgets also uses this convention)

```suggestion
bool ScopeASCOM::GetTrackingRate(TrackingRateInfo *rateInfo)
```

src/scope_ascom.h
```diff
@@ -99,6 +101,12 @@ class ScopeASCOM : public Scope
     MOVE_RESULT Guide(GUIDE_DIRECTION direction, int durationMs) override;
 
     double GetDeclinationRadians() override;
+    std::vector<Scope::TrackingRateInfo> EnumerateTrackingRates() override;
+    bool GetTracking(bool *tracking) override;
+    bool SetTracking(bool tracking) override;
+    bool CanSetTracking() override;
+    bool GetTrackingRate(TrackingRateInfo& rateInfo) override;
```
agalasso:
```suggestion
    bool GetTrackingRate(TrackingRateInfo *rateInfo) override;
```

agalasso (commented):
looks pretty good, a few minor suggestions

src/scope.cpp
```diff
@@ -1810,6 +1814,40 @@ double Scope::GetDeclinationRadians()
...
+}
+
+bool Scope::CanSetTracking()
+{
+    return false;
+}
+
+bool Scope::GetTrackingRate(TrackingRateInfo *rateInfo)
+{
+    rateInfo = &m_supportedTrackingRates.front(); // sidereal
```
agalasso: this won't actually do anything -- it just reassigns the local variable with no visible effect for the caller
Here's what we need:
```suggestion
    *rateInfo = m_supportedTrackingRates[0]; // sidereal
```

src/scope_ascom.cpp
```diff
@@ -840,6 +929,177 @@ double ScopeASCOM::GetDeclinationRadians()
...
+        case rateLunar:
+            rateInfo->name = _("Lunar");
+            break;
+        case rateSolar:
+            rateInfo->name = _("Solar");
+            break;
+        case rateKing:
+            rateInfo->name = _("King");
+            break;
+        }
```
agalasso: Let's explicitly check for an unknown rate rather than just quietly fall through. Also, I think it will be better to only update the result when the method succeeds -- no side-effects if the method fails.

```suggestion
        wxString name;
        switch (vRes.iVal)
        {
        case rateSidereal:
            name = _("Sidereal");
            break;
        case rateLunar:
            name = _("Lunar");
            break;
        case rateSolar:
            name = _("Solar");
            break;
        case rateKing:
            name = _("King");
            break;
        default:
            throw ERROR_INFO(wxString::Format("Unknown tracking rate (%d)", vRes.iVal));
        }
        rateInfo->numericalID = (TrackingRate) vRes.iVal;
        rateInfo->name = name;
```

agalasso (commented):
getting very close :)

bwdev01:
Yes, getting close - about like Christmas... <LOL>

agalasso (approved):
looks good!

PR #1459, opened 2026-08-16
Add Connect wrapper methods for other gear types besides GuideCamera

agalasso:
Add Connect wrapper methods for other gear types besides GuideCamera

Provide a place to add code common to connecting each gear type,
common to all gear subclasses.

This will the place to enumerate scope tracking rates after connecting
the scope.


bwdev01 (approved):
Looks good

PR #1464, opened 2026-09-05
Gear simulator ssg

bwdev01:


src/gear_simulator.cpp
```diff
@@ -563,14 +597,20 @@ struct SimCamState
...
-    bool ReadNextImage(usImage& img, const wxRect& subframe);
-# endif
+    void CloseDir();
+    bool ReadFitImage(usImage& img, wxString& filename, const wxRect& subframe);
 
     void Initialize();
     void FillImage(usImage& img, const wxRect& subframe, int exptime, int gain, int offset);
+    // File-related methods used for SSG
+    bool LoadNonFitsImage(cv::Mat& img, bool preProcess, wxSize& fullSize);
+    bool LoadFitsImage(usImage& img, cv::Mat& matImage, bool preProcess, wxSize& fullsize);
```
agalasso, line 610: Can we combine these into a single LoadImage method that does everything and remove the filename extension check from the caller (it's duplicated in the method and needn't be done twice).
We can also remove the preProcess argument since it is always true.

src/gear_simulator.cpp
```diff
@@ -563,14 +597,20 @@ struct SimCamState
...
+    void PreProcessImage(cv::Mat& img);
+    void GetSimDisplacements(double *pDeltaX, double *pDeltaY, double *pGearTime, bool TransformToCameraCoords);
+    bool ApplySimDisplacements(usImage *img, cv::Mat& matImage);
```
agalasso, line 613: `const` since it does not modify matImage
```suggestion
    bool ApplySimDisplacements(usImage *img, const cv::Mat& matImage);
```

src/gear_simulator.cpp
```diff
@@ -116,6 +131,8 @@ double SimCamParams::comet_rate_x;
 double SimCamParams::comet_rate_y;
 bool SimCamParams::allow_async_st4 = true;
 unsigned int SimCamParams::frame_download_ms; // frame download time, ms
+unsigned int SimCamParams::SimFileIndex = 1;
+wxString SimCamParams::SimFileTemplate = _("C:\\Temp\\phd2\\sim_image.png");
```
agalasso: not translated!

PR #1465, opened 2026-09-07
Update Czech translations in messages.po

Buloh:
Updated Czech translations for various Alpaca camera and telescope setup messages in messages.po.


agalasso (approved):
Thanks!

PR #1466, opened 2026-09-10
fix crash when some long cam-id copy

278261631:
Under Windows, the x86-compiled binary crashes due to an out-of-bounds copy when cameras with relatively long IDs are connected.
now switched to using CAMERA_ID_LENGTH as the length definition (currently=64), and it has been compiled and passed testing.

my@qhyccd.com


agalasso (approved):
thanks for the fix

PR #1469, opened 2026-09-22
Moravian cameras: update to SDK version 5.0

agalasso:
Moravian cameras: update to SDK version 5.0

User-tested with a C1-500A and with my C1-1500


bwdev01 (approved):
Does this relate to the field problem you're chasing?

